PR #14, opened 2016-10-23
Added argument to create() to allow for client options

bryangruneberg:
For my project I needed the crawler to follow redirects. 

With this change I can initialise the crawler as follows

`Crawler::create([RequestOptions::ALLOW_REDIRECTS => true])
      ->setCrawlObserver($kriechenCrawler)
      ->setCrawlProfile($kriechenCrawler)
      ->startCrawling($url);
`

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -36,14 +36,17 @@ class Crawler
     protected $crawlProfile;
 
     /**
+     * @param array $clientOptions
      * @return static
      */
-    public static function create()
+    public static function create(array $clientOptions = array())
     {
-        $client = new Client([
+        $options = array_merge([
             RequestOptions::ALLOW_REDIRECTS => false,
             RequestOptions::COOKIES         => true,
-        ]);
+        ], $clientOptions);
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for your contribution.

I'm not a fan of passing the option to the create method. Could please replace this by a fluent setter? Also add a test to prove that the change works.

bryangruneberg:
Hey there,

Sure happy to do that. 

I'll revert the changes to the constructor, and add a fluent setter that sets a request options array. Then I'll alter the $client->request() call to use the options. I think that is a cleaner way of going about it.

In terms of the test, would you like to specifically test the "allow_redirects" flag, or to test the fact that the guzzle client request() call accepts the parameters?

Thanks!

danhunsaker:
Generally, you only want to test your own library, not the dependencies.  So the Guzzle client should be mocked out anyway.

freekmurze:
I've added support for this in `v2` of the package, which will be released rather soonish.

PR #31, opened 2016-12-02
typos

blueclock:
just a couple of typos and a word order change.

@freekmurze I'll probably be using V2 in a small project that I'll show on the 14th. So our talks should dovetail nicely.


freekmurze:
Thanks! Looking forward to the 14th!

Issue #47, opened 2016-12-28
Guzzle Config Options not honored fully


mallardduck:
Whoops, sorry for the initially empty issue report. Here's what I meant to include......

------------

I've started playing with this library recently and noticed a few odd things happening with Guzzle options. From what I'm seeing in Crawler.php there are 2 bugs causing this odd behaviour. I could be wrong on these, but from my debugging the current behaviour is errant. 

## 1. Issue with create method logic

Check out [this line](https://github.com/spatie/crawler/blob/master/src/Crawler.php#L42) and line [40 (right above)](https://github.com/spatie/crawler/blob/master/src/Crawler.php#L40).

Since the function call initialises an empty array no matter what, the null coalesce will trigger as false. The default code provided will never be triggered from what I can see; effectively dead code unless I'm mistaken.

If I'm correct that the default options should still be possible then I suggest this:

```
        $client = new Client($clientOptions ?? [
                RequestOptions::COOKIES => true,
            ]);
```
Be changed to:

```
        $client = new Client( (bool) count($clientOptions) ?  $clientOptions : [
                RequestOptions::COOKIES => true,
            ]);
```

## 2. With the update to use Pools none of the client options will ever be used

Since all crawls now use the pool system for Guzzle this means that on [line 122](https://github.com/spatie/crawler/blob/master/src/Crawler.php#L122) we override any user input for options.

Unless there's a reason for this to be done, I would suggest this be updated to use:
> $this->client->getOptions()

...to populate pool options. This way the library will honour either the default settings, or user input, all the way through it's usage of guzzle.

## 3. (Optional) If both are to be addressed...

...then the client creation may need to look like:

```
        $client = new Client( (bool) count($clientOptions) ?  $clientOptions : [
                RequestOptions::CONNECT_TIMEOUT => 10,
                RequestOptions::TIMEOUT => 10,
                RequestOptions::ALLOW_REDIRECTS => false,
                RequestOptions::COOKIES => true,
            ]);
```

-----

So I'm looking forward to some feedback on this; I would have submitted a PR but I'm not 100% sure what is intended here. Once it's been discussed I'd be happy to submit anything as a PR.

freekmurze:
You're right that the passing of options seems to been have coded up a bit sloppy. Sorry for that.

Both issues should be addressed. I welcome a PR that fixes this.

mallardduck:
Awesome, just wanted to make sure I was on the right track with my assumptions!
I can definitely submit a PR to address all that.

[...]

PR #48, opened 2016-12-29
Honor user provided Guzzle opts with Pool

mallardduck:
This PR should address issue #47 that I reported the other day.

Let me know if you have any questions or concerns! I thought of a handful of ways this could be addressed, but this seemed the best. Let me know if you'd like any adjustments to how this is done.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -44,7 +44,7 @@ class Crawler
                 RequestOptions::COOKIES => true,
                 RequestOptions::CONNECT_TIMEOUT => 10,
                 RequestOptions::TIMEOUT => 10,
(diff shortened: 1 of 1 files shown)


freekmurze:
Hi, this seems like a good solution. I know the passing of the options is not easy to test automatically.

Did you run the crawler manually to test the changes?

mallardduck:
Yeah, I've done some pretty extensive manual testing for this change. I can confirm this will honor user provided Guzzle options when setting up the pool.

When no options are provided in the Crawler::create() method then it will default to using the defaults I have set. The defaults I used are based on merging the defaults I found originally being set.

freekmurze:
Cool, thank you very much!

PR #49, opened 2016-12-30
Add logic to track DOM/HTML elements...

mallardduck:
...this is super useful for tracking link redirects that could be affecting SEO.

In particular this can be useful when it comes to generating a report, or even a "patch" file, of the redirecting links.

For example you crawl a website and are looking for onsite redirects, the same (or very similar) links are on multiple pages. Once the report is done it's much easier to address the various links and pages if you have context about what page the links on and where it is. By tracking the actual link tag this is much easier; additionally a 'corrected' link can also be presented in the report.

Example of the output:

```
SEO Warning: Redirect found on page: https://grandmascookieblog.com/
This HTML link:

<a href="/cloud.php">Onsite test</a>

redirects to: https://www.liquidweb.com/cloudsites/
The final page/URL is: https://www.liquidweb.com/cloudsites/. The corrected link is:

<a href="https://www.liquidweb.com/cloudsites/">Onsite test</a>
```

### DIFF
--- a/src/CrawlObserver.php
+++ b/src/CrawlObserver.php
@@ -7,22 +7,21 @@ interface CrawlObserver
     /**
      * Called when the crawler will crawl the url.
      *
-     * @param \Spatie\Crawler\Url $url
+     * @param \Spatie\Crawler\CrawlUrl $url
      *
      * @return void
      */
-    public function willCrawl(Url $url);
+    public function willCrawl(CrawlUrl $url);
 
     /**
      * Called when the crawler has crawled the given url.
      *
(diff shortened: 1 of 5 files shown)


freekmurze:
Thank you for this, but as it contains many breaking changes, I'm holding this one off for now. I'll revisit this when my own work on our seo-tool is progressing.

mallardduck:
Sounds like a plan! I was actually just thinking the changes over again and found a better way to implement it. Much less breakage the way I have it now.

I'll keep playing with the idea on my fork until you're ready for v3, or when the seo-tool has progressed! :D

PR #50, opened 2016-12-30
Add test for User Guzzle options

mallardduck:
With this new test it added PHPunit would only pass if Guzzle's pool uses the same options provided as input.

@freekmurze your comment about tests got me curious to setup some unit tests that use options!


freekmurze:
Going to pass on this for now.

PR #51, opened 2017-01-24
Implement CrawlProfile interface for internal links

ilazaridis:
This is related to https://github.com/spatie/http-status-check/issues/22

### DIFF
--- a/src/CrawlInternalUrls.php
+++ b/src/CrawlInternalUrls.php
@@ -5,10 +5,12 @@ namespace Spatie\Crawler;
 class CrawlInternalUrls implements CrawlProfile
 {
     protected $host = '';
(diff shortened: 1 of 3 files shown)


freekmurze:
Could you add a test to prove that this works?

ilazaridis:
When I try to run existing tests adding an external link too, crawler seems to ignore this link, so I didn't go any further with tests. Is that intentional ? I didn't have time to look into it yet.

ilazaridis:
Argh, find it! Missing trailing slash from the testing url led me to wrong conclusions. I 've fixed that and added a proper test.

freekmurze:
Great PR, thanks!

PR #54, opened 2017-02-28
Patch path-relative URL crawling

### LINKED ISSUE #52: relative urls bugs!
The normalizeUrl() function in Crawler.php has not been implemented properly for the reason of :

A relative url in a page has infinite absolute url representation in relative-to-absolute transformation according to RFC1808. It varies according to current page url.

For example:
The current url is:
http://www.example.com/a/b.html,
and there is anchor:
href="c.html",
According to current normalizeUrl function, the anchor will be normalize to url:
http://www.example.com/c.html,
But it's wrong, it should be:
http://www.example.com/a/c.html

AlexVanderbist:
Check whether urls on the crawled page are path-relative urls. If so; add the directory of the current path before their path-relative path.

For example crawling this page:

```html
<!-- http://example.com/directory/subdirectory/index.html -->

<a href="example.html">example</a>
```

will add `http://example.com/directory/subdirectory/example.html` to the crawl queue instead of `http://example.com/example.html`.

Fixes #52

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -191,8 +191,8 @@ class Crawler
             ->filter(function (Url $url) {
                 return $url->hasCrawlableScheme();
             })
-            ->map(function (Url $url) {
-                return $this->normalizeUrl($url);
+            ->map(function (Url $url) use ($foundOnUrl) {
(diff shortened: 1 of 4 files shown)


freekmurze:
Great work, thanks!

PR #55, opened 2017-03-03
Use links() function to create absolute links

barryvdh:
Instead of creating your own absolute handling, there is links method on the crawler. In combination with setting the current Url, it will do the hard stuff for you.

You could probably also remove the segments and directory methods if you want.

(Disclaimer, did not actually test this, only using the tests for reference)


freekmurze:
This solution is obviously much cleaner, thanks!

I'll test it out and probably merge it in soon.

freekmurze:
@AlexVanderbist tested out your changes on some sites, and this refactor seems to work well.

Thank you!

PR #56, opened 2017-03-03
Use UriInterface instead of URL

barryvdh:
I got a bit carried away with #55 and cleaning up the Url class and felt that you could probably entirely remove your own Url implementation and use the UriInterface + Uri implementation from guzzle/psr7.

That will do that hard stuff for parsing urls etc. Guzzle also uses that object for links.
Combined with the links() resolving from #55, there is not real need for it anymore.

As you can see, tests are hardly changed (mostly `Url::create()` -> `new Uri()` and `$uri->getHost()` instead of `$uri->host` and immutability; `$this->uri = $this->uri->withPort()` instead of `$this->uri->setPort()`). Only small tweaks, which I will point out in comments.

But I understand that you cannot just remove a public class and change interfaces with SemVer in minor versions, so I'll just leave this here for future inspiration ;)
(Note, still not manually tested, only with the testserver)


src/Crawler.php
```diff
@@ -178,68 +181,52 @@ protected function getCrawlRequests(): Generator
 
             $this->crawlQueue->markAsProcessed($crawlUrl);
 
-            yield new Request('GET', (string) $crawlUrl->url);
+            yield new Request('GET', $crawlUrl->url);
```
barryvdh: Guzzle uses PSR-7, so accepts actual UriInterface objects

src/Crawler.php
```diff
@@ -178,68 +181,52 @@ protected function getCrawlRequests(): Generator
...
+    protected function addAllLinksToCrawlQueue(string $html, UriInterface $foundOnUrl)
     {
-        $allLinks = $this->extractAllLinks($html);
+        $allLinks = $this->extractAllLinks($html, $foundOnUrl);
 
         collect($allLinks)
-            ->filter(function (Url $url) {
-                return $url->hasCrawlableScheme();
+            ->filter(function (UriInterface $url) {
+                return in_array($url->getScheme(), ['http', 'https']);
```
barryvdh: Removed a simple convenience method

src/CrawlQueue.php
```diff
@@ -98,7 +99,7 @@ public function removeProcessedUrlsFromPending()
     protected function contains(Collection $collection, CrawlUrl $searchCrawlUrl): bool
     {
         foreach ($collection as $crawlUrl) {
-            if ($crawlUrl->url->isEqual($searchCrawlUrl->url)) {
+            if ((string) $crawlUrl->url === (string) $searchCrawlUrl->url) {
```
barryvdh: Removed a simple convenience method

src/Crawler.php
```diff
@@ -178,68 +181,52 @@ protected function getCrawlRequests(): Generator
...
     {
-        $domCrawler = new DomCrawler($html);
+        $domCrawler = new DomCrawler($html, $foundOnUrl);
 
-        return collect($domCrawler->filterXpath('//a')->extract(['href']))
-            ->map(function ($url) {
-                return Url::create($url);
+        return collect($domCrawler->filterXpath('//a')->links())
+            ->map(function (Link $link) {
+                return new Uri($link->getUri());
```
barryvdh: Use Symfony DomCrawler to generate the URL

src/Crawler.php
```diff
@@ -178,68 +181,52 @@ protected function getCrawlRequests(): Generator
...
-                ->setHost($this->baseUrl->host)
-                ->setPort($this->baseUrl->port);
-        }
-
-        if ($url->isProtocolIndependent()) {
-            $url->setScheme($this->baseUrl->scheme);
-        }
-
-        return $url->removeFragment();
+        return $url->withFragment('');
```
barryvdh: Return new Uri instance without the fragment. Rest is done by Symfony/UriInterface already

tests/CrawlerTest.php
```diff
@@ -32,12 +33,12 @@ public function it_will_crawl_all_found_urls()
 
         $this->assertCrawledOnce([
             ['url' => 'http://localhost:8080/'],
-            ['url' => 'http://localhost:8080/link1', 'foundOn' => 'http://localhost:8080/'],
```
barryvdh: This is normalized by Uri

tests/UrlTest.php
```diff
@@ -1,171 +0,0 @@
-<?php
-
-namespace Spatie\Crawler\Test;
-
-use Spatie\Crawler\Url;
-
-class UrlTest extends TestCase
```
barryvdh: This is removed because the entire Uri is used from Guzzle, no need to test it ourself

freekmurze:
Thank you for this. Closing PR now, but will revisit this when working on a new major version.

freekmurze:
We're going to tag a new major version of the crawler soon, currently considering pulling this in...

barryvdh:
I'm not stopping you ;)

freekmurze:
Mmm, lots of conflicts...  If I find the time, I'll just perform the necessary changes manually on the `v3` branch.

PR #65, opened 2017-09-16
Implementing depth 

joetito1:
Implementing depth using nicmart/Tree library. As the Crawler extracts URLs from pages, it builds a Tree structure of parent/children. If the current URL has reached the maximum depth, it skips adding future URLs to the crawlQueue.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -12,6 +12,8 @@ use Symfony\Component\DomCrawler\Link;
 use Psr\Http\Message\ResponseInterface;
 use GuzzleHttp\Exception\RequestException;
 use Symfony\Component\DomCrawler\Crawler as DomCrawler;
+use Tree\Node\Node;
+use Tree\Visitor\PreOrderVisitor;
 
 class Crawler
 {
@@ -32,6 +34,13 @@ class Crawler
 
     /** @var \Spatie\Crawler\CrawlQueue */
     protected $crawlQueue;
+
+    /** @var int */
+    protected $depth;
+
+    private $tree;
+
+    private $rootNode;
 
     /**
      * @param array $clientOptions
@@ -74,6 +83,18 @@ class Crawler
         return $this;
     }
 
+	/**
+	 * @param int $depth
+	 *
+	 * @return $this
+	 */
+	public function setDepth(int $depth)
+	{
+		$this->depth = $depth;
+
+		return $this;
+	}
+
     /**
      * @param \Spatie\Crawler\CrawlObserver $crawlObserver
      *
@@ -113,7 +134,15 @@ class Crawler
 
         $this->crawlQueue->add($crawlUrl);
 
+        $this->tree = new Node((string)$baseUrl);
+
         $this->startCrawlingQueue();
+
+		$visitor = new PreOrderVisitor;
+
+		$yield = $this->tree->accept($visitor);
+		print_r($yield);
+		// $yield will contain nodes A, B, C, D, G, H, E, F
 
         $this->crawlObserver->finishedCrawling();
     }
@@ -147,6 +176,8 @@ class Crawler
             $promise->wait();
 
             $this->crawlQueue->removeProcessedUrlsFromPending();
+
+            echo '-----' . PHP_EOL;
         }
     }
 
@@ -202,6 +233,13 @@ class Crawler
                 return $this->crawlQueue->has($url);
             })
             ->each(function (Url $url) use ($foundOnUrl) {
+
+            	echo 'Found on URL: ' . $foundOnUrl . PHP_EOL;
+            	$parent = $this->findParentNode((string) $foundOnUrl);
+
+            	echo 'Found parent URL' . (string) $parent->getValue() . PHP_EOL;
+				$parent->addChild(new Node((string)$url));
+
                 $this->crawlQueue->add(
                     CrawlUrl::create($url, $foundOnUrl)
                 );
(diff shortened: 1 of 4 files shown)


danhunsaker:
So, this adds maximum depth limiting, but doesn't seem to actually implement "depth" as a whole, which I would expect to also mean including the crawl depth of each new URL to the result data, at a minimum.

src/Crawler.php
```diff
@@ -74,6 +81,18 @@ public function setConcurrency(int $concurrency)
         return $this;
     }
 
+	/**
+	 * @param int $depth
+	 *
+	 * @return $this
+	 */
+	public function setDepth(int $depth)
```
freekmurze: Could you rename this function to `setMaximumDepth` and all variables `$depth` to `maximumDepth`?

src/Crawler.php
```diff
@@ -192,7 +213,7 @@ protected function addAllLinksToCrawlQueue(string $html, Url $foundOnUrl)
             ->filter(function (Url $url) {
                 return $url->hasCrawlableScheme();
             })
-            ->map(function (Url $url) use ($foundOnUrl) {
+            ->map(function (Url $url) {
```
freekmurze: 👍

src/Crawler.php
```diff
@@ -202,9 +223,16 @@ protected function addAllLinksToCrawlQueue(string $html, Url $foundOnUrl)
                 return $this->crawlQueue->has($url);
             })
             ->each(function (Url $url) use ($foundOnUrl) {
-                $this->crawlQueue->add(
-                    CrawlUrl::create($url, $foundOnUrl)
-                );
+
+				$newNode = null;
```
freekmurze: This seems off. Why do you create a null variable? Can this be cleaned up?

src/Crawler.php
```diff
@@ -202,9 +223,16 @@ protected function addAllLinksToCrawlQueue(string $html, Url $foundOnUrl)
...
+
+            	$this->addToLinkTree($this->linkTree, (string)$url, $foundOnUrl, $newNode);
+
+				if(($this->depth == 0) || ($newNode->getDepth() <= $this->depth)) {
```
freekmurze: Use strict comparison here `===`

src/Crawler.php
```diff
@@ -227,4 +255,24 @@ protected function normalizeUrl(Url $url): Url
...
     }
+
+
+	/**
+	 * @param $node \Tree\Node\Node
+	 * @param $url string
+	 * @param $parentUrl string
+	 * @param $newNode \Tree\Node\Node
+	 */
+	protected function addToLinkTree($node, string $url, string $parentUrl, &$newNode) {
```
freekmurze: Can we avoid the `&` on `$newNode`?

src/Crawler.php
joetito1: Absolutely. This makes more sense.

src/Crawler.php
joetito1: Done!

src/Crawler.php
joetito1: So this and your other request for not using the reference & were the only way I could figure out how to return the exact node in the tree for the new node. I can definitely look for ways to make this better if you won't accept it otherwise.

joetito1:
@freekmurze Addressed the rest of your concerns with my latest commits. Got rid of the pass by reference that was used previously.

freekmurze:
Could you also update the readme? Add an example on how to use the new functionality.

joetito1:
@freekmurze Updated documentation with a description of the default functionality and an example of how to set maximumDepth

danhunsaker:
Still think it would be useful to have access to the depth each URL was added at in the crawl results. If this is already possible, it would be helpful to document that as well.

joetito1:
@danhunsaker So you mean that each Url would have a $depth value in it so that you could add logic based off that in your Observer?

danhunsaker:
Something like that. On the other hand, the entire tree structure might be useful as well. Probably both, if I was building this, though others tend to be more elegant than I am in these scenarios.

joetito1:
@danhunsaker - Yeah, that's not a bad idea. Definitely something I could add in the future if the Author thinks it would be useful.

joetito1:
@freekmurze This is my first time doing a pull request on github. Is there anything else I need to do to get this merged in?

freekmurze:
@joetito1 pretty good for a first contribution 👍 

All you need to have is a little bit of patience. When I've got a little time on my hands, I'll handle this PR further.

freekmurze:
I'm having second thoughts about this PR not on a technical level but more on a functional level. I wonder what the actual use case is for knowing the depth of a page.  

Probably you want to use it to limit the amount of pages the crawler will crawl. But in that case using depth is not a good way. Level 2 might have thousands of pages already. 

Assume that you don't want to get to for example level 4. If you put a link on your homepage to a page that is on level 4, suddenly that page and all pages will go up in the level tree.

Could you explain how knowing the depth is useful?

joetito1:
@freekmurze The goal of limiting depth is mostly to limit the number of pages crawled. I have a use case where I search websites for information that's typically available in pages that are 1-2 pages deep from the home page and without limiting the depth the crawler wastes a TON of time on things like crawling entire blog archives. These aren't pages I manage, so I can't implement any other good logic to get to these pages besides limiting depth. I totally understand that there could be thousands of links on a page 2 deep, but that doesn't happen much on legit websites.

freekmurze:
How about just limiting the amount of pages crawled. Wouldn't that be safer?

joetito1:
Not really? If I say limit the crawl to 500 pages and they had a blog with 501 posts before the page I need, I'd never get to the page(s) I need. With maximum depth limiting, I would always (eventually) get to the page(s) i'm looking for.

freekmurze:
In your use case: if even a single page on a top level contains a link to one of the archived pages, the whole archive will be crawled. Isn't that a bit fragile?

Do your archive pages have something specific in the url to recognise them by? Then you just could not crawl them specifically use a `CrawlProfile`. Feels more robust.

Thoughts?

freekmurze:
Ok, I'll merge this in after you've rebased against the current master.

joetito1:
@freekmurze All done..I think. This is making me realize how much I suck at using git :P

freekmurze:
Still some conflicts unfortunately...

joetito1:
@freekmurze There we go :)

freekmurze:
Much better, thanks!

PR #78, opened 2017-10-05
Fixed tests

BrentRobert:
Both assertCrawledOnce and assertNotCrawled only asserted the last URL because the assertion happened outside of the loop.

Only 1 test failed because of this, and was fixed by adding $this.resetLog();


freekmurze:
Thanks!

PR #83, opened 2017-11-09
Update to PHPUnit 6

gabriel-caruso:
As we support `PHP 7`, I've updated our suite of tests to `PHPUnit 6`.

PS: I won't apply [this](https://styleci.io/analyses/XNg3ZE) `StyleCI` changes. There are not related to this PR :sweat_smile:


freekmurze:
Thank you!

PR #86, opened 2017-12-08
Speed improvement. Not calculate the node depth, if maximumDepth not defined

systream:


src/Crawler.php
```diff
@@ -280,7 +280,7 @@ protected function addAllLinksToCrawlQueue(string $html, Url $foundOnUrl)
             ->each(function (Url $url) use ($foundOnUrl) {
                 $node = $this->addtoDepthTree($this->depthTree, (string) $url, $foundOnUrl);
 
-                if (! $this->shouldCrawlAtDepth($node->getDepth())) {
+                if (! $this->shouldCrawlAtDepth($node)) {
```
freekmurze: Rename this method to `shouldCrawl`. That'll better describe it's new purpose.

systream: Thanks, I forgot to rename it. 
Done.

freekmurze:
Thanks!

PR #87, opened 2017-12-08
Add change crawl queue support

systream:
Above a certain amount of link, the bottleneck is the crawl queue. So i decided, to make it changeable.
The crawler class needed some refactoring to achieve this. I hope it helps.

### DIFF
--- a/src/CollectionCrawlQueue.php
+++ b/src/CollectionCrawlQueue.php
@@ -1,0 +1,103 @@
+<?php
+
+namespace Spatie\Crawler;
+
+use Illuminate\Support\Collection;
+use Spatie\Crawler\Exception\UrlNotFoundByIndex;
+
+class CollectionCrawlQueue implements CrawlQueue
+{
+    /** @var \Illuminate\Support\Collection */
+    public $urls;
+
+    /** @var \Illuminate\Support\Collection */
+    public $pendingUrls;
+
+    public function __construct()
+    {
+        $this->urls = collect();
+
+        $this->pendingUrls = collect();
+    }
+
+    public function add(CrawlUrl $url): CrawlQueue
+    {
+        if ($this->has($url)) {
+            return $this;
+        }
+
+        $this->urls->push($url);
+
+        $url->setId($this->urls->keys()->last());
+        $this->pendingUrls->push($url);
+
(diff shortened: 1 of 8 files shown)


src/CollectionCrawlQueue.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * @param int $id
+     *
+     * @return \Spatie\Crawler\CrawlUrl|null
+     */
+    public function getUrlById(int $id): CrawlUrl
+    {
+        if (! isset($this->urls->values()[$id])) {
+            throw new UrlNotFoundByIndex(sprintf('#%d crawl url not found in collection', $id));
```
freekmurze: `sprintf` doesn't get used often in our code bases. Replace this with string interpolation.

systream: Sorry, fixed. 

I have removed the PDO based driver. I thought it could be a good example, but you are right.

src/CollectionCrawlQueue.php
```diff
@@ -0,0 +1,108 @@
...
+            if ($crawlUrl->url->isEqual($searchCrawlUrl->url)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /** @return \Spatie\Crawler\CrawlUrl|null */
+    public function getPendingUrl()
```
freekmurze: Rename this function `getFirstPendingUrl`.

src/CollectionCrawlQueue.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $first = $this->pendingUrls->first();
```
freekmurze: You can just return the result of the function: if `pendingUrls` is empty, `null` will be returned.

src/CollectionCrawlQueue.php
systream: Done

src/CollectionCrawlQueue.php
systream: I didn't know that. Fixed.

freekmurze: 👍

freekmurze:
Thanks!

PR #95, opened 2017-12-11
StyleCI fixes and add `npm install` to readme

juukie:
Before tinkering out for the V3, I had to add this ;)


freekmurze:
Thanks!

PR #96, opened 2017-12-11
[v3] Upgrade Browsershot

juukie:
Hi @freekmurze, this is what I had in mind. It allows users to set a custom created Browsershot instance via `setBrowsershot()`. What do you think?


juukie:
I'm not sure how to let Travis install puppeteer. Something like
```
before_script:
  - npm install -g puppeteer
  - ./tests/server/start_server.sh
```
?

src/Crawler.php
```diff
@@ -347,15 +345,29 @@ protected function addtoDepthTree(Node $node, string $url, string $parentUrl)
...
         }
 
-        $html = $browsershot->bodyHtml();
+        $this->browsershot = new Browsershot();
 
-        return html_entity_decode($html);
+        return $this->browsershot;
+    }
+
+    protected function setBrowsershot(Browsershot $browsershot)
```
freekmurze: Shouldn't this method be `public`?

freekmurze:
> I'm not sure how to let Travis install puppeteer and use Node 7 since that's what is required by Browsershot.

The tests for this PR seem to be running fine, no?

juukie:
Thanks for looking it over. I might also add a test to cover the setBrowsershot() method.
Visibility has changed to public.

Travis issue was solved later on so that's ok :).

juukie:
I've added the test. Although it passes, it does not really test if our custom instance is being used. 
Is there a way to check if `setBrowsershot()` really changes the $browsershot property? Since it's a protected property, I can't use `assertSame($browsershot, $crawler->browsershot)`. Any idea on this?

freekmurze:
Thanks!

PR #102, opened 2017-12-22
V3

brendt:


brendt:
@freekmurze I'm wondering whether it would be good to rename all `$url` variables to `$uri` now that we're using the PSR-7 `UriInterface`?

Feels more consistent to me, what's your opinion?

PR #108, opened 2018-01-17
Fix setting maximum response size

AlexVanderbist:
When a crawled response exceeds the set maximum response size (±2MB by default) it will not be parsed for new URLs to crawl. 

Users parsing the `$response->getBody()` in `CrawlObserver::hasBeenCrawled()` should still be cautious as `setMaximumResponseSize` doesn't stop the responses from being passed to `hasBeenCrawled()`

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -3,23 +3,23 @@
 namespace Spatie\Crawler;
 
 use Generator;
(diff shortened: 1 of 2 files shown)


freekmurze:
Perfect!

PR #116, opened 2018-02-20
fix to mark url's that should not be crawled as processed

SERFF:
In my use case the while loop goes into an endless loop, when marked as processed this is no longer an issue.


freekmurze:
Thanks!

PR #118, opened 2018-02-22
Update travis config

TVke:


tests/NoSandboxCrawler.php
```diff
@@ -0,0 +1,152 @@
...
+use InvalidArgumentException;
+use Psr\Http\Message\UriInterface;
+use Spatie\Crawler\CrawlSubdomains;
+use Symfony\Component\DomCrawler\Link;
+use Psr\Http\Message\ResponseInterface;
+use Tightenco\Collect\Support\Collection;
+use GuzzleHttp\Exception\RequestException;
+use Symfony\Component\DomCrawler\Crawler as DomCrawler;
+
+class NoSandboxCrawler extends Crawler
```
freekmurze: Even though technically this makes the tests pass, this solution is not maintainable.

If we change something in the crawler class, there's a high change that we'll forget to update it here.

Instead of coping all the code, we should introduce a way to make the browsershot instance configurable from the outside.

First thought would be to add let the `executeJavaScript` function optionally accept an instance or Browsershot or a closure where in a user can configure an instance of browsershot.

That way we can can configure Browsershot in our tests to our hearts content and we can remove all this whole `NoSandboxCrawler` class.

tests/CrawlerTest.php
```diff
@@ -76,7 +76,7 @@ public function it_can_crawl_uris_without_scheme()
     /** @test */
     public function it_can_crawl_all_links_rendered_by_javascript()
     {
-        $browsershot = new Browsershot();
+        $browsershot = new Browsershot;
```
freekmurze: We always use `()` when newing up stuff. Please add them back.

freekmurze:
Thanks!

PR #136, opened 2018-03-21
Fix typos and grammer

TVke:


src/Crawler.php
```diff
@@ -302,7 +303,7 @@ protected function convertBodyToString(StreamInterface $bodyStream, $readMaximum
 
     /**
      * @param ResponseInterface|null $response
-     * @param CrawlUrl $crawlUrl
+     * @param CrawlUrl               $crawlUrl
```
freekmurze: Remove this change.

src/Crawler.php
```diff
@@ -317,7 +318,7 @@ protected function handleCrawled(ResponseInterface $response, CrawlUrl $crawlUrl
 
     /**
      * @param RequestException $exception
-     * @param CrawlUrl $crawlUrl
+     * @param CrawlUrl         $crawlUrl
```
freekmurze: Remove this change

freekmurze:
Thanks.

PR #138, opened 2018-03-22
Removed php native type hinting of url, queue and crawler pool Closures

Redominus:
In order to support any kind of url indexing, int type hinting should be removed. Also, added phpdoc comments where needed.

### DIFF
--- a/src/CrawlQueue/CollectionCrawlQueue.php
+++ b/src/CrawlQueue/CollectionCrawlQueue.php
@@ -40,11 +40,11 @@ class CollectionCrawlQueue implements CrawlQueue
     }
 
     /**
-     * @param int $id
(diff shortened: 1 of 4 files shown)


danielstgt:
This PR opens quite a lot of possibilities, e.g. an implementation of a md5 hash based queue. That would result in very performant lookups for large sites, no need for loops.

If this PR gets accepted and there is an interest, I could make a PR with the hash based queue (which would also remove the need for the collections dependency).

However, thanks to [PHP's Type Juggling](http://php.net/manual/en/language.types.type-juggling.php) there are absolutely no drawbacks at all with the mixed type for the current queue suggestion.

freekmurze:
Thanks!

PR #139, opened 2018-03-23
Added meta robots follow/nofollow directive check and ignoring

Redominus:
This change allows to honor robots meta tag nofollow directive. It can be enabled and disabled with to methods. Its is checked in every page after loading JS.


freekmurze:
Thanks for your work on this, but we'll soon work on implementing this ourselves via https://github.com/spatie/robots-txt

PR #141, opened 2018-04-16
Depth tree logic called only when needed.

Redominus:
If there is no depth restriction we can save A LOT of execution time by not building and checking the depthTree.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -422,13 +422,14 @@ class Crawler
                 return $this->crawlQueue->has($url);
             })
             ->each(function (UriInterface $url) use ($foundOnUrl) {
-                $node = $this->addtoDepthTree($this->depthTree, $url, $foundOnUrl);
-
-                if (strpos($url->getPath(), '/tel:') === 0) {
-                    return;
+                if ($this->maximumDepth !== null) {
+                    $node = $this->addtoDepthTree($this->depthTree, $url, $foundOnUrl);
+                    if (! $this->shouldCrawl($node)) {
+                        return;
+                    }
                 }
 
-                if (! $this->shouldCrawl($node)) {
+                if (strpos($url->getPath(), '/tel:') === 0) {
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


src/Crawler.php
```diff
@@ -371,13 +371,14 @@ protected function addAllLinksToCrawlQueue(string $html, UriInterface $foundOnUr
                 return $this->crawlQueue->has($url);
             })
             ->each(function (UriInterface $url) use ($foundOnUrl) {
-                $node = $this->addtoDepthTree($this->depthTree, $url, $foundOnUrl);
-
-                if (strpos($url->getPath(), '/tel:') === 0) {
-                    return;
+                if ($this->maximumDepth !== null) {
```
freekmurze: Change this to 

```
if (is_null($this-> maximumDepth) {
   return;
}
```

so we use an indentation level in the rest of the function.

Redominus: with that change all $url would return; because maximunDepth never changes.
I've anohter idea, changing shoulCrawl signature to
```php
/**
* @param Node|null $node
*/
protected function shouldCrawl( $node): bool
```
and placing at the begining of addtoDepthTree a is_null check
```php
protected function addtoDepthTree(Node $node, UriInterface $url, UriInterface $parentUrl)
    {
        $returnNode = null;
if (is_null($this-> maximumDepth)) {
return $returnNode;
}
```

Redominus: Another idea, changing shouldCrawl signature so it can call addtoDepthTree

freekmurze: > with that change all $url would return; because maximunDepth never changes.
I don't understand your argument.

cmizzi: Why prefer `is_null` instead of `=== null` ? It prevents function call and with a lots of call, this can reduce function call overhead, no ?

freekmurze: Avoiding `is_null` is a micro optimization. We use `is_null` in other packages so i'd like to keep it consistent.

My main gripe was just reversing the condition so we can lose a level of indentation.

Redominus: Just making that change would have prevented to add any url to the crawl queue. Instead I moved the the check inside addToDepthTree and now returns a Node without checking or adding it to the depth tree.

freekmurze:
Meanwhile I've implemented this myself. Thanks for your work on this.

Issue #114, opened 2018-01-29
Crawler should respect robots.txt


freekmurze:
Right now the crawler checks all links regardless of what's in the `robots.txt` file. Ideally the crawler should have a `respectRobotsTxt` method that parses the `robots.txt` file (or header) and only crawl the allowed links.

I'd accept a PR (with tests) that adds this to the package.

danhunsaker:
Would it be acceptable to add a dependency on an external library that handles the parsing? If so, this would probably be a quick PR to put together, but if not, it seems important to know that in advance.

freekmurze:
> Would it be acceptable to add a dependency on an external library that handles the parsing?

It would certainly be acceptable if that dependency has tests, good documentation and is maintained.

brendt:
@danhunsaker Any specific packages coming to mind?

danhunsaker:
Not particularly (I tend to do things in Laravel, so L5-specific packages are where my familiarity lies), but `webignition/robots-txt-file` seems solid. Others are simpler to use, but haven't been updated as recently, or don't have (enough) tests.

sebastiandedeyne:
Don't forget there's a meta tag for robots values too

brendt:
I'm working on this!

Redominus:
meta tag for robtos already addressed in pull request  #139

PR #145, opened 2018-05-04
Respect robots

brendt:
Addresses #114

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,16 @@ This package comes with three `CrawlProfiles` out of the box:
 - `CrawlInternalUrls`: this profile will only crawl the internal urls on the pages of a host.
 - `CrawlSubdomainUrls`: this profile will only crawl the internal urls and its subdomains on the pages of a host.
 
+### Ignoring robots.txt and robots meta
+
+By default, the crawler will respect robots data. It is possible to disable these checks like so:
+
+```php
+Crawler::create()
+    ->ignoreRobots()
+    ...
+```
+
 ## Setting the number of concurrent requests
 
 To improve the speed of the crawl the package concurrently crawls 10 urls by default. If you want to change that number you can use the `setConcurrency` method.
--- a/README.md
+++ b/README.md
@@ -165,6 +165,11 @@ Crawler::create()
     ...
 ```
 
+Robots data can come from either a `robots.txt` file, meta tags or response headers.
+More information on the spec can be found here: [http://www.robotstxt.org/](http://www.robotstxt.org/).
+
+Parsing robots data is done by our package [spatie/robots-txt](https://github.com/spatie/robots-txt).
+
 ## Setting the number of concurrent requests
 
 To improve the speed of the crawl the package concurrently crawls 10 urls by default. If you want to change that number you can use the `setConcurrency` method.
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -56,7 +56,7 @@ class Crawler
     protected $maximumDepth = null;
 
     /** @var bool */
-    protected $ignoreRobots = false;
+    protected $respectRobots = true;
 
     /** @var \Tree\Node\Node */
     protected $depthTree;
@@ -159,7 +159,7 @@ class Crawler
      */
     public function ignoreRobots()
     {
-        $this->ignoreRobots = true;
+        $this->respectRobots = false;
 
         return $this;
     }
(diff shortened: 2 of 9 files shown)


src/Crawler.php
```diff
@@ -61,6 +68,9 @@ class Crawler
     /** @var Browsershot */
     protected $browsershot = null;
 
+    /** @var \Spatie\Robots\RobotsTxt */
+    private $robotsTxt = null;
```
freekmurze: make this `protected`

src/Crawler.php
```diff
@@ -145,6 +155,18 @@ public function setMaximumDepth(int $maximumDepth)
         return $this;
     }
 
+    /**
+     * @param bool $ignoreRobots
+     *
+     * @return $this
+     */
+    public function ignoreRobots(bool $ignoreRobots = true)
```
freekmurze: Do not accept a boolean here, let's add a dedicated `respectRobots` method.

src/Crawler.php
```diff
@@ -301,9 +338,14 @@ protected function convertBodyToString(StreamInterface $bodyStream, $readMaximum
...
 
+    protected function createRobotsTxt(UriInterface $uri): RobotsTxt
+    {
+        return RobotsTxt::create($uri->withPath('/robots.txt'));
+    }
+
     /**
      * @param ResponseInterface|null $response
-     * @param CrawlUrl $crawlUrl
+     * @param CrawlUrl               $crawlUrl
```
freekmurze: I don't like this indentation, but I guess StyleCI will take care of this.

src/Crawler.php
```diff
@@ -393,6 +435,10 @@ protected function addAllLinksToCrawlQueue(string $html, UriInterface $foundOnUr
 
     protected function shouldCrawl(Node $node): bool
     {
+        if (! $this->ignoreRobots && ! $this->robotsTxt->allows($node->getValue())) {
```
freekmurze: Remove `&&` by using an extra `if`

src/Crawler.php
```diff
@@ -505,4 +551,22 @@ protected function maximumCrawlCountReached(): bool
...
         return $this->crawledUrlCount >= $this->maximumCrawlCount;
     }
+
+    protected function mayIndex(RobotsHeaders  $robotsHeaders, RobotsMeta $robotsMeta): bool
+    {
+        if ($this->ignoreRobots) {
+            return true;
+        }
+
+        return $robotsHeaders->mayIndex() && $robotsMeta->mayIndex();
```
freekmurze: Avoid `&&` but using an extra `if`

src/Crawler.php
```diff
@@ -238,6 +260,8 @@ public function startCrawling($baseUrl)
 
         $crawlUrl = CrawlUrl::create($this->baseUrl);
 
+        $this->robotsTxt = $this->createRobotsTxt($crawlUrl->url);
```
freekmurze: Will this work when the given url is not the root of a domain?
eg `https://example.com/page/subpage`

README.md
```diff
@@ -155,6 +155,16 @@ This package comes with three `CrawlProfiles` out of the box:
 - `CrawlInternalUrls`: this profile will only crawl the internal urls on the pages of a host.
 - `CrawlSubdomainUrls`: this profile will only crawl the internal urls and its subdomains on the pages of a host.
 
+### Ignoring robots.txt and robots meta
+
+By default, the crawler will respect robots data. It is possible to disable these checks like so:
```
freekmurze: Add some extra info here on what a robots.txt and header is? Also mention that the actual check is being done by our robots-txt package.

src/Crawler.php
brendt: Yes, the URL is used in `createRobotsTxt` like so:

```
$uri->withPath('/robots.txt')
```

By the robots spec, the `robots.txt` must always be in the root.

src/Crawler.php
brendt: I've disabled it in my auto formatter :)

src/Crawler.php
```diff
@@ -52,6 +55,9 @@ class Crawler
     /** @var int|null */
     protected $maximumDepth = null;
 
+    /** @var bool */
+    protected $ignoreRobots = false;
```
freekmurze: I think the code will be clearer if we use something positive here. Let's rename it to `respectRobots` and set it to `true`.

README.md
brendt: Also done.

brendt:
@freekmurze I fixed the failing tests and think it's ready to merge now. One question though: how will we go about testing this for a while without tagging a release?

PR #148, opened 2018-05-15
Allow the crawler to run without any observers

sebastiandedeyne:
Useful for features like https://github.com/spatie/laravel-responsecache/issues/133

What do you think @freekmurze?

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -32,7 +32,7 @@ class Crawler
     protected $baseUrl;
 
     /** @var array[\Spatie\Crawler\CrawlObserver] */
-    protected $crawlObservers;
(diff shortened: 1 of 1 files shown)


sebastiandedeyne:
Would remove the need for line 36 here

https://github.com/sebastiandedeyne/sebastiandedeyne.com/blob/48244edf97921aff86848d88f206ca3f3568697b/app/Console/Commands/WarmCommand.php#L34-L37

freekmurze:
Pretty elegant solution 👍

brendt:
This will also be fixed in my refactor PR: https://github.com/spatie/crawler/pull/150/files#diff-d0975b110122f222a14d71c183e72026R101

Issue #143, opened 2018-05-02
Provide Browsershot object or bodyHtml to CrawlObserver


kloor:
Currently, Browsershot executes Javascript on a page after that page has been provided to a CrawlObserver. It would be useful if instead that execution could happen first, and either the Browsershot object or the bodyHtml from that object could be provided in some way to the observer.

brendt:
This one should be fixed in 4.15: https://github.com/spatie/crawler/releases/tag/4.1.5

Issue #142, opened 2018-05-02
Crawler parses body for all status codes


kloor:
The addAllLinksToCrawlQueue method is run on all HTTP responses including response codes like 301 and 302 for redirects. This isn't a noticeable issue when not executing Javascript, as the bodies are usually empty. When executing Javascript it becomes a problem as Puppeteer follows the redirects, resulting in links on the destination page being parsed even if that page is excluded by the CrawlProfile.

It would be nice if response codes could be whitelisted (maybe 200 or 2XX) or blacklisted (maybe 301, 302, 307 and 308) before Javascript is executed and the body parsed.

kloor:
Being able to specify my own CrawlRequestFulfilled handler is quite nice, and did allow me to solve this issue in my own code. Thanks!

PR #150, opened 2018-05-15
Code refactor

brendt:
I was working on #142 and #143 when I noticed that I was adding more and more logic in the same functions, making them very long. I also did this when I recently added robots.txt check, and I feel we can do better.

This PR proposes a first, simple, refactor. I've moved the logic to handle a request's success or failure to separate classes, each having their own dependencies. There are a lot more things that can be split, and the dependencies given to both handlers can be optimised.

I think a review at this point is necessary though, to make sure we're on the right track. If the PR as it is now is merged, it will already result in a cleaner solution for the issues mentioned above. I can also spend more time to further structure and split the code better. Th result would be a simpler to navigate codebase, where it's simpler for newer devs of this package, like me, to know what happens where, instead of the huge class right now.

Anyways, feedback is appreciated.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -12,16 +12,11 @@ use GuzzleHttp\Psr7\Uri;
 use GuzzleHttp\Psr7\Request;
 use Spatie\Robots\RobotsTxt;
 use InvalidArgumentException;
-use Spatie\Robots\RobotsMeta;
 use GuzzleHttp\RequestOptions;
-use Spatie\Robots\RobotsHeaders;
 use Psr\Http\Message\UriInterface;
 use Spatie\Browsershot\Browsershot;
-use Psr\Http\Message\StreamInterface;
 use Symfony\Component\DomCrawler\Link;
-use Psr\Http\Message\ResponseInterface;
 use Spatie\Crawler\CrawlQueue\CrawlQueue;
-use GuzzleHttp\Exception\RequestException;
 use Spatie\Crawler\CrawlQueue\CollectionCrawlQueue;
 use Symfony\Component\DomCrawler\Crawler as DomCrawler;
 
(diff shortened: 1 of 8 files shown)


src/Crawler.php
```diff
@@ -284,50 +237,20 @@ public function startCrawling($baseUrl)
 
     protected function startCrawlingQueue()
     {
+        $fulfilledHandler = $this->getFulfilledHandler();
```
freekmurze: Rename this `onFulfill` and use it immediately in the pool. Same for the `failedHandler`, rename that to `onFail`

src/Crawler.php
```diff
@@ -404,7 +288,7 @@ protected function getCrawlRequests(): Generator
         }
     }
 
-    protected function addAllLinksToCrawlQueue(string $html, UriInterface $foundOnUrl)
+    public function addAllLinksToCrawlQueue(string $html, UriInterface $foundOnUrl)
```
freekmurze: Why is this public?

src/Crawler.php
```diff
@@ -561,37 +443,24 @@ protected function maximumCrawlCountReached(): bool
...
-        if (! $robotsHeaders->mayIndex()) {
-            return false;
-        }
-
-        if (! $robotsMeta->mayIndex()) {
-            return false;
-        }
-
-        return true;
+        return new CrawlRequestFulfilled(
```
freekmurze: I'd inject only the crawler itself here instead of all these. That way we don't have to touch this function when extra stuff needs to be passed to `CrawlRequestFulfilled` in the future.

src/LinkAdder.php
```diff
@@ -0,0 +1,115 @@
...
+                }
+
+                $crawlUrl = CrawlUrl::create($url, $foundOnUrl);
+
+                $this->crawler->addToCrawlQueue($crawlUrl);
+            });
+    }
+
+    /**
+     * @param string                         $html
```
freekmurze: I don't like this formatting

src/LinkAdder.php
```diff
@@ -0,0 +1,115 @@
...
+     * @param \Psr\Http\Message\UriInterface $foundOnUrl
+     *
+     * @return \Illuminate\Support\Collection|\Tightenco\Collect\Support\Collection|null
+     */
+    protected function extractLinksFromHtml(string $html, UriInterface $foundOnUrl)
```
freekmurze: Can we use a nullable typehint here?

brendt: I don't think so, because either `Illuminate\Support\Collection` or `Tightenco\Collect\Support\Collection` is returned.

src/CrawlerProperties.php
```diff
@@ -0,0 +1,180 @@
...
+namespace Spatie\Crawler;
+
+use Spatie\Robots\RobotsTxt;
+use Psr\Http\Message\UriInterface;
+use Spatie\Browsershot\Browsershot;
+use Spatie\Crawler\CrawlQueue\CrawlQueue;
+
+trait CrawlerProperties
+{
+    public function setConcurrency(int $concurrency): self
```
sebastiandedeyne: Isn't a `self` return type problematic when extending a class? I know you'd rather not have this issue, but we should be consistent when using `protected`...

src/Handlers/CrawlRequestFulfilled.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->linkAdder = new LinkAdder($this->crawler);
+    }
+
+    public function __invoke(ResponseInterface $response, $index)
+    {
+        $crawlUrl = $this->crawler->getCrawlQueue()->getUrlById($index);
+
+        $body = $this->convertBodyToString($response->getBody(), $this->crawler->getMaximumResponseSize());
+
+        $robotsHeaders = RobotsHeaders::create($response->getHeaders());
```
sebastiandedeyne: This function feels cluttered with `RobotsHeaders` & `RobotsMeta` implementation details. Maybe we should create a local `Robots` class (created from a `ResponseInterface`) to deal with these?

src/LinkAdder.php
```diff
@@ -0,0 +1,115 @@
...
+        $allLinks = $this->extractLinksFromHtml($html, $foundOnUrl);
+
+        collect($allLinks)
+            ->filter(function (UriInterface $url) {
+                return $this->hasCrawlableScheme($url);
+            })
+            ->map(function (UriInterface $url) {
+                return $this->normalizeUrl($url);
+            })
+            ->each(function (UriInterface $url) use ($foundOnUrl) {
```
sebastiandedeyne: These early returns could be refactored to a `filter` chain. Sounds cleaner to me, but might look a but messy in practice.

src/Handlers/CrawlRequestFulfilled.php
brendt: Good idea, I'll do that.

src/CrawlerProperties.php
brendt: You're right, let's discuss this further in person though.

src/LinkAdder.php
brendt: Seems cleaner, indeed. Though the last check on maximumCrawlCountReached() must be in the each loop. I've refactored the rest though.

sebastiandedeyne: To clarify intent there, you could do `->take($this->cawler->remainingCrawlCount())`

freekmurze:
Very nice refactor 👍

PR #158, opened 2018-06-01
Changed access method for $defaultClientOptions

mattiabasone:
Changed from self to static for overwriting $defaultClientOptions in a derived class


freekmurze:
Thanks!

PR #168, opened 2018-06-15
Improve extendablity of Crawler

damiantw:
I recently upgraded from version 4.1.0 of this package. I had created my own crawler class that extended the Spatie\Crawler\Crawler class with a few functions overridden to meet my business logic needs (Specifically relating to recording URLs that could not be crawled due to robots or a crawl profile). 

Refactorings in recent releases of this package make overriding these functions no longer possible. This pull request improves the extendability of the Crawler class by allowing the end user to provide custom handler classes for fulfilled and failed requests. Abstract classes for these handlers have been added that the default and user provided handlers must extend.

In addition, functions with a return type declarations of 'self' have been replaced with a return type of 'Crawler' such that these functions can be overridden.


src/Handlers/CrawlRequestFulfilled.php
```diff
@@ -2,29 +2,12 @@
...
-use Spatie\Crawler\CrawlUrl;
-use Spatie\Crawler\LinkAdder;
+use Psr\Http\Message\ResponseInterface;
 use Spatie\Crawler\CrawlerRobots;
 use Spatie\Crawler\CrawlSubdomains;
-use Psr\Http\Message\StreamInterface;
-use Psr\Http\Message\ResponseInterface;
 
-class CrawlRequestFulfilled
+class CrawlRequestFulfilled extends CrawlRequestFulfilledAbstract
```
freekmurze: I really don't like the `Abstract` suffix. Rename this class `CrawlRequestFulfilled` and add a prefix `Default` to the actual implementation.

src/Crawler.php
```diff
@@ -338,8 +382,8 @@ protected function startCrawlingQueue()
             $pool = new Pool($this->client, $this->getCrawlRequests(), [
                 'concurrency' => $this->concurrency,
                 'options' => $this->client->getConfig(),
-                'fulfilled' => new CrawlRequestFulfilled($this),
-                'rejected' => new CrawlRequestFailed($this),
+                'fulfilled' => $this->getCrawlFulfilledHandler(),
```
freekmurze: Let's drop the `get` prefix and rename this to `crawlFulfilledHandler`

damiantw:
I've pushed the requested changes and additional tests.

freekmurze:
Thanks!

brendt:
A friendly FYI, you can extend `self`, but should replace `self` with the concrete class name in subclasses. You can see `self` as a token which gets replaced at runtime by the current class.

https://3v4l.org/VuE7b

PR #175, opened 2018-07-01
Corrected not marking processed links on a falsy shouldCrawl return

raaaahman:
I had an issue with creating CrawlProfiles and even using those povided:

Problem was when the shouldCrawl function returned false, the crawling loop skipped the markProcessed method call, so the link would be checked again, which causes php to exceed maximum runtime.

It is working fine on my project now.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -407,6 +407,8 @@ class Crawler
                 continue;
             }
 
+	$this->crawlQueue->markAsProcessed($crawlUrl);
+
             if ($this->crawlQueue->hasAlreadyBeenProcessed($crawlUrl)) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


freekmurze:
The tests seems to fail for this one...

raaaahman:
Hum, sorry. I'll see what I can do, but it'll take me some time.

raaaahman:
Hi, 
I've been running the tests on my side and I have 15 failures with a message like: 

> Did not find hasBeenCrawled: http://localhost:8080/
 exactly one time in the log but 0 times

or

>Did not find hasBeenCrawled: http://localhost:8080/header-disallow - found on http://localhost:8080/

Care to explain what it means?
Thanks,

PR #177, opened 2018-07-28
Moves the number of urls counted to the Queue

peterfox:
I've been playing around with the Crawler package lately but noticed the number of URLs processed tallied up in the crawler instance and it seems to me that it would make more sense being in the Queue applied to the crawler that stores all the urls that have been added. This isn't a breaking change other than for those who have their own Queue implementations but it should be a minor change to add a count.

The advantage of this is it means that a shared Queue between crawlers will stop when the queue has globally reached any count set between all of them instead of when that individual crawler reached the maximum crawl count. I could imagine there being scenarios where someone wants either though and I don't mind putting in the work if that is considered a more useful approach?

### DIFF
--- a/src/CrawlQueue/CollectionCrawlQueue.php
+++ b/src/CrawlQueue/CollectionCrawlQueue.php
@@ -90,6 +90,11 @@ class CollectionCrawlQueue implements CrawlQueue
         return $this->pendingUrls->first();
     }
 
+    public function count(): int
+    {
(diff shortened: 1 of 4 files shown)


freekmurze:
Thank you for this. Adding an extra method to the `CrawlQueue` interface will break all custom queues. So I'm not going to merge this in now, but will revisit it when creating a new major version.

PR #189, opened 2018-10-26
Add support for a customizable delay between every page crawled

mattiasgeniar:


tests/CrawlerTest.php
```diff
@@ -353,6 +353,28 @@ public function it_should_not_follow_nofollow_links()
         $this->assertNotCrawled([['url' => 'http://localhost:8080/nofollow', 'foundOn' => 'http://localhost:8080/']]);
     }
 
+    /** @test */
+    public function it_respects_the_requested_delay_between_requests()
```
freekmurze: Very nice test, well done!

freekmurze:
Thank you for this.

Even though the true way to achieve this would be to use Guzzle middleware, I've decided to merge it in, so it's much easier to use a delay.

PR #192, opened 2018-11-07
Randomized Proxy Support

jonrobinson:
To effectively scrape certain sites that limit requests per IP I think the package should support randomized proxies. Since Guzzle only allows you to specify one proxy per client this is achieved at the `Pool` level, updating the request options (adding a random proxy) on each pool. 

Another caveat is that a very large pool will still have the same IP address sending requests in quick succession. To allow for the greatest flexibility here we've also add a `poolItemLimit` parameter that limits the pool size -- allowing you to rotate IP addresses down to each request. 

These are easily used in two functions:

**Setting Random Proxies Per Pool**

```
$crawler->setProxies([
    'ips' => '123.45.67.89,123.45.67.88,123.45.67.87...',
    'username' => 'laravel',
    'password' => 'secret',
    'port' => '12343',
])
```

**Limiting 1 Item Per Pool**

```
$crawler->setPoolItemLimit(1)

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -335,6 +335,14 @@ class Crawler
         return $this->baseUrl;
     }
 
+    public function setProxies(array $proxyConfig)
+    {
+        $this->proxyConfig = $proxyConfig;
+        $this->usingProxies = true;
+        
+        return $this;
+    }
+
     /**
      * @param \Psr\Http\Message\UriInterface|string $baseUrl
      */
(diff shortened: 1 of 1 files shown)


Redominus:
You already can use an arbitrary number of proxies persisted anyway you want.
Passing GuzzleHttp client options in the crawler create metod.  In the client options you can pass any kind of middleware that handles proxies, user agents and anything you want.

PR #206, opened 2019-02-23
Documentation fix

jpedryc:
It seems that one of the CrawlProfiles is *CrawlSubdomains* not *CrawlSubdomainUrls*


freekmurze:
Thanks!

PR #225, opened 2019-05-21
The Crawler is handling relative paths after redirects incorrectly.

Redominus:
Hi!
I've discovered the crawler is handling incorrectly relative paths after redirects. When the redirect happens, all links are extracted as if the foundOn was the original url instead of the last url. This solution will try to take advantage of guzzle's X-Guzzle-Redirect-History header.

### DIFF
--- a/src/Handlers/CrawlRequestFulfilled.php
+++ b/src/Handlers/CrawlRequestFulfilled.php
@@ -3,11 +3,11 @@
 namespace Spatie\Crawler\Handlers;
 
 use GuzzleHttp\Psr7\Uri;
-use GuzzleHttp\RedirectMiddleware;
 use Spatie\Crawler\Crawler;
 use Spatie\Crawler\CrawlUrl;
 use Spatie\Crawler\LinkAdder;
 use Spatie\Crawler\CrawlerRobots;
+use GuzzleHttp\RedirectMiddleware;
 use Psr\Http\Message\UriInterface;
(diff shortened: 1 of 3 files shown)


src/Handlers/CrawlRequestFulfilled.php
```diff
@@ -2,10 +2,12 @@
 
 namespace Spatie\Crawler\Handlers;
 
+use GuzzleHttp\Psr7\Uri;
 use Spatie\Crawler\Crawler;
 use Spatie\Crawler\CrawlUrl;
 use Spatie\Crawler\LinkAdder;
```
Redominus: I don't like this import ordering method. :disappointed:

danhunsaker: Alphabetical is the correct order for PSR-2 compliant code, so that's what gets enforced.

Redominus: I thought order by length was being used because no styleci errors adns watching the original file:
https://github.com/spatie/crawler/blob/e991498972a2e41daac0f12ba47346c2b9922ac2/src/Handlers/CrawlRequestFulfilled.php#L5-L13

I will fix it tomorrow ASAP.

Redominus:
Looks like StyleCI wants Its downstairs of imports.

riasvdv:
Thanks!

PR #227, opened 2019-06-05
ADD A Crawl Resume Logic

zigwartybund:
It would be awesome to add the ability to resume crawl after a shutdown. Maybe with a database logic to store crawl queue...and mark out witch links have already been crawled....


freekmurze:
This PR is unreviewable. Please open a PR with only your changes and without conflicts.

PR #235, opened 2019-06-20
CollectionCrawlQueue::has() does not accept string

BenMorel:
If `$crawlUrl` is not an instance of `CrawlUrl`, `CrawlUrl::create()` is used, and it only accepts a `UriInterface`.
So the actual allowed types are `CrawlUrl|UriInterface`.

### DIFF
--- a/src/CrawlQueue/CollectionCrawlQueue.php
+++ b/src/CrawlQueue/CollectionCrawlQueue.php
@@ -77,7 +77,7 @@ class CollectionCrawlQueue implements CrawlQueue
     }
 
     /**
-     * @param CrawlUrl|\Psr\Http\Message\UriInterface|string $crawlUrl
+     * @param CrawlUrl|\Psr\Http\Message\UriInterface $crawlUrl
      *
      * @return bool
      */


brendt:
I think we should support all three. Are you able to change this PR accordingly? If not I'll take a look at it.

I'd do something like

```php
    public function has($crawlUrl): bool
    {
        if (is_string($crawlUrl)) {
            $crawlUrl = CrawlUrl::create(new UriInterface($crawlUrl));
        } elseif ($crawlUrl instanceof UriInterface) {
            $crawlUrl = CrawlUrl::create($crawlUrl);
        }
```

brendt:
Furthermore, let's also move this logic to `CrawlUrl::make`, which can take a `string`, `UriInterface` or a `CrawlUrl`, and return a new `CrawlUrl` object.

BenMorel:
No worries, I can update the PR accordingly.

Note that IMHO, it would be actually better to restrict the allowed types, rather than expanding them. This allows to use type-hints to enforce types, catch errors earlier, etc.
For example this method could just accept a `UriInterface` and the caller could just do `->has($crawlUrl->url)`, this is not much effort and makes the overall thing cleaner if you ask me. But this would be a BC break and would require a version bump, so it's out of the question for now I guess.

That being said, no part in the code currently sends a `string`, so expanding the list of types is not useful right now, are you sure you want to procede this way?

I'll do either way, I just want to be sure that we're going in the right direction here.

brendt:
You know what? You're right. My reasoning was that we couldn't simply drop support for strings, because it was documented as "the intended behaviour". But since this never worked in the first place, there's no need to support it now and remove it in the next major version. Thanks for thinking along :)

I'll merge this PR as is, and we'll revisit it also in the next major version to only accept `UriInterface` or `CrawlUrl` — we should think about which one.

Thanks!

brendt:
Tagged as 4.4.2: https://github.com/spatie/crawler/releases/tag/4.4.2

BenMorel:
Thank you! 👍

BTW, how open are you to preparing a v5 release? I'm planning to use this library quite extensively in the near future, and foresee a good number of such improvements that would require a BC break and PHP 7.1+ compatibility. I could send some PRs if you're interested!

brendt:
Always open for PRs. If there are worthwhile features that require a major version bump, I'll be happy to do that. 

We do try to bundle breaking changes though, so it might take a little longer to have the release actually tagged

Issue #233, opened 2019-06-19
How to implement retry?


BenMorel:
First of all thanks for this high-quality, well thought out package 👍 

I'm crawling a website, and am getting a lot of failures (I'm probably getting rate-limited).

The documentation does not explain how to handle retries of failed URLs:

- is it handled automatically? Is it configurable?
- if not, what's the best way to implement it? Should I manually call `Crawler::addToCrawlQueue()` inside `CrawlObserver::crawlFailed()` (in this case I'll need a way to keep a counter of retries for every URL, to give up after n retries)

A pointer in the right direction would be highly appreciated!

A note in the documentation would also be nice for other users who might have the same question (I can PR it if you wish after I get your feedback and get to play with it).

Redominus:
Retry logic can be implemented in the guzzle handler stack using middleware. In fact, guzzlehttp comes with a static retry middleware builder(example below!).
First read the [docs](http://docs.guzzlephp.org/en/stable/handlers-and-middleware.html) about guzzle handlers and middleware.

Then try this simple exmple code:

1. Declare two callables to handle when to retry logic and retry delay calculationj logic: ``

1.

Redominus:
Retry logic can be implemented in the guzzle handler stack using middleware. In fact, guzzlehttp comes with a static retry middleware builder(example below!).
First read the docs about guzzle handlers and middleware.

Then try this simple exmple code:

1. Declare two callables to handle retry logic and retry delay calculation: 
```php
function retryDecider()
{
    return function (
        $retries,
        Request $request,
        Response $response = null,
        RequestException $exception = null
    ) {
        // Limit the number of retries to 5
        if ($retries >= 5) {
            return false;
        }

        // Retry connection exceptions
        if ($exception instanceof ConnectException) {
            return true;
        }

        if ($response) {
            // Retry on server errors
            if ($response->getStatusCode() >= 500) {
                return true;
            }
        }
        return false;
    };
}

/**
 * delay 1s 2s 3s 4s 5s
 *
 * @return Closure
 */

function retryDelay()
{
    return function ($numberOfRetries) {
        return 1000 * $numberOfRetries;
    };
}
```

2. Push to the handler the retry middleware with the logic.
```php 
$handler->push(
        Middleware::retry(
            retryDecider(),
            retryDelay()
        )
    );

BenMorel:
@Redominus thanks! I did not think the retry logic belonged to Guzzle.

[...]

PR #237, opened 2019-06-20
Implement retries

BenMorel:
This is a working, **backwards-compatible** implementation of the retry feature suggested in #233.

Example usage:

```php
$retryProfile = new class implements \Spatie\Crawler\RetryProfile
{
    public function shouldRetry(CrawlUrl $crawlUrl, RequestException $exception): bool
    {
        $response = $exception->getResponse();

        // only retry network errors and 5xx server errors
        if ($response !== null && $response->getStatusCode() < 500) {
            return false;
        }

        // maximum 5 attempts
        return $crawlUrl->getAttempts() < 5;
    }
};

Crawler::create($options)
    ->setRetryProfile($retryProfile)
    ->startCrawling($url);
```

### Why this feature?

As suggested by @Redominus in the related issue, one could use Guzzle's retry middleware instead. The issue is that this middleware performs consecutive retries **on the same URL**, leaving no time for any temporary glitch to resolve itself.

### What does this do instead?

If the `RetryProfile` in use instructs it to, the Crawler will push the failed URL back to the pending queue. The failed URL will be processed **after other pending URLs**, leaving more time between attempts, and avoiding a single URL to be considered a failure straight away in case the target server is down for a short amount of time.

This will also be useful when using a persistent crawl queue, such as a database-backed queue: the crawler can be interrupted at any time, and the failed URLs will persist in the pending queue.

This makes retries a first-class citizen, part of Crawler's core.

### What about backwards compatibility?

BC is fully preserved. The only tricky part was to extend the `CrawlQueue` interface, which misses a method to push a URL back to the pending queue.

The solution found was to extend it: `RetryableCrawlQueue` extends `CrawlQueue`, adding the method to support this feature. `CollectionCrawlQueue` now implements `RetryableCrawlQueue`, supporting the retry feature out-of-the-box.

For existing implementations of `CrawlQueue` that may be existing in the wild, they will continue to work as they did. If someone using a custom `CrawlQueue` attempts to make use of a `RetryProfile` without updating their crawl queue implementation, retries will just not be performed.

The `RetryProfile` docblock makes this very clear.

### Future scope

When BC can be broken (v5), `RetryableCrawlQueue` can be merged into `CrawlQueue`.
A `@todo` was left in the source code to remember this.

---

@Redominus @freekmurze @brendt Looking forward to your feedback! 👍

### DIFF
--- a/src/CrawlQueue/CollectionCrawlQueue.php
+++ b/src/CrawlQueue/CollectionCrawlQueue.php
@@ -5,7 +5,7 @@ namespace Spatie\Crawler\CrawlQueue;
 use Spatie\Crawler\CrawlUrl;
 use Spatie\Crawler\Exception\UrlNotFoundByIndex;
 
-class CollectionCrawlQueue implements CrawlQueue
+class CollectionCrawlQueue implements RetryableCrawlQueue
 {
     /** @var \Illuminate\Support\Collection|\Tightenco\Collect\Support\Collection */
     protected $urls;
@@ -115,5 +115,23 @@ class CollectionCrawlQueue implements CrawlQueue
         }
 
         return false;
+    }
+
+    /**
+     * @param CrawlUrl $crawlUrl
+     *
+     * @return void
+     */
+    public function retry(CrawlUrl $crawlUrl)
+    {
+        if (! $this->contains($this->urls, $crawlUrl)) {
+            return;
+        }
+
+        if ($this->contains($this->pendingUrls, $crawlUrl)) {
+            return;
+        }
+
+        $this->pendingUrls->push($crawlUrl);
     }
 }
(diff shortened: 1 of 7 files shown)


Redominus:
In my opinion this belongs to an observer.

freekmurze:
Thank you for your work on this but I'm going to pass on this for now. Believe it or not, but I'd like to keep this package as simple as it can be. For my projects a Guzzle middleware like [this one](https://github.com/caseyamcl/guzzle_retry_middleware) works fine, so I don't need to maintain retry functionality myself.

BenMorel:
I understand. I'll maintain my own fork if this feature becomes critical for me. Thanks for your feedback.

PR #239, opened 2019-06-21
Fix $delayBetweenRequests type, add doc

BenMorel:
`$delayBetweenRequests` is only set from within `setDelayBetweenRequests()`, which is type-hinted as `int`. So this property can never be a `float`.

`getDelayBetweenRequests()` should not implicitly convert it to a `float`, as it's only used by `usleep()`, which accepts an `int`.

This yields a useless implicit conversion chain from `int` => `float` => `int`.

I've also added a doc to state that it accepts a number of *milliseconds*, which is otherwise not documented.

Note that *this is a small BC break*, in case someone overrides the non-final `getDelayBetweenRequests()` method - which is unlikely, but still, I'm not sure whether this is acceptable for you to release in a minor or patch version.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -71,7 +71,7 @@ class Crawler
     /** @var string */
     protected $crawlRequestFailedClass;
 
-    /** @var float */
+    /** @var int */
     protected $delayBetweenRequests = 0;
 
     /** @var   */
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #244, opened 2019-06-27
[v5] Get rid of CrawlUrl::$id

BenMorel:
Guzzle's Pool allows string keys, allowing us to use the URL as the key, and avoiding the need for a numeric ID in `CrawlUrl`.

*This PR targets `master` for now, but I think the project should be branched to `v5` and this PR retargeted.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to `spatie/crawler` will be documented in this file.
 
+## Unreleased (5.0.0)
+
+Breaking changes:
+
+- exception `UrlNotFoundByIndex` has been renamed to `UrlNotFound`
+- method `CrawlQueue::getUrlById()` has been removed
+- new method `CrawlQueue::get()` has been added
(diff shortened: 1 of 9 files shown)


BenMorel:
Note that while I removed `CrawlQueue::getUrlById()`, I had to add a `get()` method to get the `CrawlUrl` object back from the stringified URL, as Guzzle accepts strings but not objects as keys:

https://github.com/spatie/crawler/blob/6fed8417f0f9489c7a391ffa0cc1460778cc351e/src/CrawlQueue/CrawlQueue.php#L15

I hope to be able to further simplify the code by adding support for object keys in the Guzzle's `Pool` at some point. I opened an issue with Guzzle, this should be fairly easy to do, provided that it can be done in a BC manner:

https://github.com/guzzle/guzzle/issues/2322

freekmurze:
I've created a `v5-wip` branch and retargeted the PR.

brendt:
Looks great, thanks Ben!

BenMorel:
Thank you, I'll soon move forward with another PR!

PR #246, opened 2019-07-10
Feature/set user agent

denvers:
I've added new functionality in order to set a custom User Agent and let Crawler respect custom User Agent rules from robots.txt according to it. I'd like to hear any feedback in order to get it merged :-)

I also followed your CONTRIBUTING guidelines at https://github.com/spatie/crawler/blob/master/CONTRIBUTING.md

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -74,6 +74,9 @@ class Crawler
     /** @var int */
     protected $delayBetweenRequests = 0;
 
+    /** @var string */
+    protected $userAgent = "*";
+
     /** @var   */
     protected static $defaultClientOptions = [
         RequestOptions::COOKIES => true,
@@ -311,6 +314,13 @@ class Crawler
         return $this;
     }
 
+    public function setUserAgent(string $userAgent): Crawler
+    {
+        $this->userAgent = $userAgent;
+
+        return $this;
+    }
+
     public function getBrowsershot(): Browsershot
     {
         if (! $this->browsershot) {
@@ -348,7 +358,7 @@ class Crawler
 
         $this->robotsTxt = $this->createRobotsTxt($crawlUrl->url);
 
-        if ($this->robotsTxt->allows((string) $crawlUrl->url) ||
+        if ($this->robotsTxt->allows((string) $crawlUrl->url, (string) $this->userAgent) ||
             ! $this->respectRobots
         ) {
             $this->addToCrawlQueue($crawlUrl);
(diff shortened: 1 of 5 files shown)


andrzejkupczyk:
I may be wrong but this approach is faulty. Imagine this is how someone initializes Crawler:
```php
$configuration = [
    'headers' => [
        'user-agent' => 'Whatever/1.0.0',
    ],
];

Crawler::create($configuration)->setUserAgent('my-agent');
```

In this case which User Agent will be used? From what I see the `setUserAgent()` method does not change HTTP client configuration.

denvers:
Hi @andrzejkupczyk , thanks for your comment.

In this case the `$configuration` is only passed to `GuzzleHttp/Client`. The `setUserAgent` method is sets the user-agent which is used by `Crawler`. These won't conflict, but I agree with you this can be confusing.

I can detect if user-agent is set through `$configuration` to give it to `setUserAgent`, so you know those user-agents are the same. Except when you use `setUserAgent` after `Crawler` (and thus) `GuzzleHttp/Client` is initialised. In that specific case the user-agents could differ from each other. This won't affect the `Crawler` / `Robots` behaviour as `Client` is not involved in this detection mechanism.

danhunsaker:
The problem here is that if the UA reported to the server differs from the UA used to confirm access, you're violating the standard. Truthfully, `Crawler` should be pulling the UA from the `Client` as needed, rather than maintaining its own value. Generally speaking, the UA should also be static, not changing between requests or processes in the first place, so the correct mechanism is, and always will be, to properly configure the UA in the `Client` and then leave it alone from there. If you really need to change up UAs for whatever reason, it's easy enough to generate a second `Client`, and set up a second `Crawler` to use it.

brendt:
I agree with @danhunsaker that we need to look at the actual UA from the client, as this is the only real source of truth.

Nevertheless, it's good to have `Crawler::setUserAgent`, but it should modify the client underneath, and not store a copy of the UA.

src/Crawler.php
```diff
@@ -74,6 +74,9 @@ class Crawler
     /** @var int */
     protected $delayBetweenRequests = 0;
 
+    /** @var string */
+    protected $userAgent = '*';
```
brendt: Always use and modify the client's UA instead of managing a copy in the `Crawler` class.

denvers: Fixed it. It now uses the client configuration as single source of truth and modifies the client user-agent with setUserAgent(). Added several tests also to verify this functionality.

brendt:
Released as 4.5.0: https://github.com/spatie/crawler/releases/tag/4.5.0

brendt:
Thank you @denvers !

PR #249, opened 2019-08-04
Fix bugs related to changing the user agent header

andrzejkupczyk:
This fixes some issues introduced in the 4.5.0 release (https://github.com/spatie/crawler/pull/246).
Header names are [not case sensitive](https://tools.ietf.org/html/rfc7230#section-3.2) and yet `Notice: Undefined index: User-Agent` occurs when trying to crawl using Crawler created with configuration such as:
```php
Crawler::create(['headers' => ['user-agent' => 'WhatEver/1.0']]);
```

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -20,6 +20,8 @@ use Spatie\Crawler\Exception\InvalidCrawlRequestHandler;
 
 class Crawler
 {
+    public const DEFAULT_USER_AGENT = 'SpatieCrawler/4.5';
+
     /** @var \GuzzleHttp\Client */
     protected $client;
 
@@ -80,6 +82,9 @@ class Crawler
         RequestOptions::CONNECT_TIMEOUT => 10,
         RequestOptions::TIMEOUT => 10,
         RequestOptions::ALLOW_REDIRECTS => false,
+        RequestOptions::HEADERS => [
+            'User-Agent' => self::DEFAULT_USER_AGENT,
+        ],
     ];
 
     public static function create(array $clientOptions = []): Crawler
@@ -314,7 +319,11 @@ class Crawler
     public function setUserAgent(string $userAgent): Crawler
     {
         $clientOptions = $this->client->getConfig();
-        $clientOptions['headers']['User-Agent'] = strtolower($userAgent);
+
+        $headers = array_change_key_case($clientOptions['headers']);
+        $headers['user-agent'] = $userAgent;
+
+        $clientOptions['headers'] = $headers;
 
         $this->client = new Client($clientOptions);
 
@@ -323,7 +332,15 @@ class Crawler
 
     public function getUserAgent(): string
     {
-        return $this->client->getConfig('headers')['User-Agent'];
+        $headers = $this->client->getConfig('headers');
+
+        foreach (array_keys($headers) as $name) {
+            if (strtolower($name) === 'user-agent') {
+                return (string) $headers[$name];
+            }
+        }
+
+        return static::DEFAULT_USER_AGENT;
     }
 
     public function getBrowsershot(): Browsershot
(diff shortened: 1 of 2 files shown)


src/Crawler.php
```diff
@@ -323,7 +332,15 @@ public function setUserAgent(string $userAgent): Crawler
...
-        return $this->client->getConfig('headers')['User-Agent'];
+        $headers = $this->client->getConfig('headers');
+
+        foreach (array_keys($headers) as $name) {
+            if (strtolower($name) === 'user-agent') {
+                return (string) $headers[$name];
+            }
+        }
+
+        return static::DEFAULT_USER_AGENT;
```
brendt: If we didn't need `DEFAULT_USER_AGENT` previously, couldn't we simply throw an exception here? No user agent in the client's config seems like something we don't need a graceful fallback for.

src/Crawler.php
```diff
@@ -314,7 +319,11 @@ public function setBrowsershot(Browsershot $browsershot)
     public function setUserAgent(string $userAgent): Crawler
     {
         $clientOptions = $this->client->getConfig();
-        $clientOptions['headers']['User-Agent'] = strtolower($userAgent);
+
+        $headers = array_change_key_case($clientOptions['headers']);
```
brendt: Why do we need this, if we're already doing lowercase checks in `getUserAgent`?

andrzejkupczyk: This is necessary because we want to be sure that we are adding a new header or updating one that already exists. Otherwise there is a risk that we will duplicate it and which one should be returned by the `getUserAgent` method then? This is not too obvious in my opinion. Guzzle has similar implementation https://github.com/guzzle/guzzle/blob/master/src/Client.php#L196

src/Crawler.php
andrzejkupczyk: Why should we throw an exception if the user agent header is not required? This would break the `LinkAdder::shouldCrawl()` method. Better solution is to change return type declaration to `?string` and return `null` when there is no user agent. Do you agree?

brendt: You're right, let's return `*`

src/Crawler.php
brendt: Doesn't that mean that this check is redundant? However you construct a client:

```php
$this->client = new Client($clientOptions);
```

Guzzle will make internally sure there will never be two `user-agent` headers because of the check you linked (https://github.com/guzzle/guzzle/blob/master/src/Client.php#L198)

andrzejkupczyk: You may be right, though this should be unit tested. I'll take care of it.

brendt: Awesome, thanks!

andrzejkupczyk: @brendt, I'm afraid that without changing array keys case, the following test:
```php
public function it_should_ignore_user_agents_header_case()
{
    $clientConfig = ['headers' => ['user-agent' => 'foo']];
    $newUserAgent = 'bar';

    $crawler = Crawler::create($clientConfig)->setUserAgent($newUserAgent);
    $actualUserAgent = $crawler->getUserAgent();

    $this->assertSame($newUserAgent, $actualUserAgent);
}
```
fails with message:
```
Failed asserting that two strings are identical.
Expected :'bar'
Actual   :'foo'
```

src/Crawler.php
andrzejkupczyk: Instead of removing constant I've only changed it's value to avoid magic string.

brendt:
Looking good, thanks!

src/Crawler.php
brendt: No problem, we'll merge it like this

brendt:
Tagged as 4.6.2: https://github.com/spatie/crawler/releases/tag/4.6.2

Issue #243, opened 2019-06-27
Roadmap for v5


BenMorel:
I'm creating this issue to keep track of the tasks for v5.

## New features

- [ ] proper support for persistent (database-backed) queues (#137, #200), allowing to perform a crawl in discrete steps at different points in time, and resume an interrupted crawl; note that this would benefit a lot from #237 (putting a failed URL back onto the persistent crawl queue) if you're willing to revisit this
- [ ] automatically set the `Referer` header
- [ ] (to be discussed) ship a PDO implementation of a database-backed crawl queue, which would also serve as a reference implementation

## Fixes

- [ ] fix the performance issues (#210, #241, and as a result, #236) (positive side effect: removing one dependency)

## Code improvements

- [X] Get rid of `$id` in `CrawlUrl`: Guzzle's `Pool` [can yield string keys](https://github.com/guzzle/guzzle/pull/1203) as well, we do not need an int, and can use the URL as the key to greatly simplify the code - **implemented** in #244
- [ ] Improve performance of `CollectionCrawlQueue` by using an indexed array rather than [looping through the URLs](https://github.com/spatie/crawler/blob/67974a810f6a342c333c4e348fcfad7c1375ffde/src/CrawlQueue/CollectionCrawlQueue.php#L103-L117)
- [ ] Add return types wherever possible (example: `CrawlQueue::getFirstPendingUrl() : ?CrawlUrl`)
- [ ] Change signature of `CrawlQueue::has` to `(CrawlUrl $crawlUrl)` to be consistent with the other methods
- [ ] Decouple `CrawlRequestFailed|FulfilledClass` from Guzzle by replacing it with an interface; this will allow better static analysis than a invokable class name, and make the contract more explicit, such as: `public function crawlFailed(RequestException $exception, CrawlUrl $url)`
- [ ] Make `CrawlProfile` an interface instead of an abstract class

## Cleanup

- [ ] remove deprecated methods (`Crawler::endsWith()`)
- [ ] If possible, get rid of `CrawlQueue::get()`, introduced in #244 as a temporary workaround for Guzzle's inability to use objects as keys

brendt:
Thanks Ben for the effort you've already put into this, much appreciated 👏

You mentioned that a `v5` branch would still be helpful, there's a `v5-wip`, which I just rebased on master. I'll try my best to keep it up to date, and suggest you target that branch.

brendt:
Also, let me know when you need input or want help with anything.

[...]

PR #250, opened 2019-08-06
Faster default crawl queue implementation

BenMorel:
🐌 **The issue with the collection crawl queue**

The default crawl queue is currently based on Illuminate Collections, and suffers from many flaws that make it slow: it [loops through entire collections to find objects, and makes ultra-heavy use of `Uri`-to-string conversions](https://github.com/spatie/crawler/blob/c5a351bf1859dfdbe46dac67bc22f26759de00c3/src/CrawlQueue/CollectionCrawlQueue.php#L111-L115):

```php
foreach ($collection as $crawlUrl) {
    if ((string) $crawlUrl->url === (string) $searchCrawlUrl->url) {
        return true;
    }
}
```

This implementation works fine for crawling small websites (< 1000 pages); after that, the crawler starts to be CPU-bound (~100% CPU usage) and starts slowing down.

After a few thousand pages, it almost comes to a halt, and takes something like 15 seconds to move from one page to another. This delay raises exponentially with the number of pages in the queue.

:rocket: **NEW: the array crawl queue**

This PR brings a 100% compatible implementation:

- based on native arrays; no more Collection overhead (wrapping, copying, filtering);
- indexed by URL: we can instantly find a given URL with a lookup, without having to loop through the collection;
- `CrawlUrl` ids are strings, using the full URL and no more artificial integer keys to lookup;
- Shortcuts: no more `contains()`, we can now use `isset()` directly.

💨 **Is is faster?**

It *is* faster. Much, **MUCH** faster. When the original implementation almost came to a halt after a few thousand pages, the new implementation is still performing 10 requests per second after 40,000 pages, and is not even CPU bound.

**That's what I suggest we make this new crawl queue the default right away, and deprecate the other one.**

---

This PR is 100% backwards compatible, and as such does not need to wait for [v5](https://github.com/spatie/crawler/issues/243).

In v5, we might want to remove the dependency on Illuminate Collections entirely. This was a noble idea, but `Collection` is overkill for this use case.

### DIFF
--- a/src/CrawlQueue/ArrayCrawlQueue.php
+++ b/src/CrawlQueue/ArrayCrawlQueue.php
@@ -2,10 +2,10 @@
 
 namespace Spatie\Crawler\CrawlQueue;
 
-use Psr\Http\Message\UriInterface;
-use Spatie\Crawler\CrawlUrl;
-use Spatie\Crawler\Exception\UrlNotFoundByIndex;
 use TypeError;
+use Spatie\Crawler\CrawlUrl;
+use Psr\Http\Message\UriInterface;
+use Spatie\Crawler\Exception\UrlNotFoundByIndex;
 
 /**
  * Crawl queue implemented with arrays.
@@ -77,7 +77,7 @@ class ArrayCrawlQueue implements CrawlQueue
     {
         if ($crawlUrl instanceof CrawlUrl) {
             $url = (string) $crawlUrl->url;
-        } else if ($crawlUrl instanceof UriInterface) {
+        } elseif ($crawlUrl instanceof UriInterface) {
             $url = (string) $crawlUrl;
         } else {
             throw new TypeError(sprintf(
--- a/src/CrawlQueue/ArrayCrawlQueue.php
+++ b/src/CrawlQueue/ArrayCrawlQueue.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace Spatie\Crawler\CrawlQueue;
+
+use Psr\Http\Message\UriInterface;
+use Spatie\Crawler\CrawlUrl;
+use Spatie\Crawler\Exception\UrlNotFoundByIndex;
+use TypeError;
+
+/**
+ * Crawl queue implemented with arrays.
+ */
+class ArrayCrawlQueue implements CrawlQueue
+{
+    /**
+     * All known URLs, indexed by URL string.
+     *
+     * @var CrawlUrl[]
+     */
+    protected $urls = [];
+
+    /**
+     * Pending URLs, indexed by URL string.
+     *
+     * @var CrawlUrl[]
+     */
+    protected $pendingUrls = [];
+
+    public function add(CrawlUrl $url) : CrawlQueue
+    {
+        $urlString = (string) $url->url;
+
+        if (! isset($this->urls[$urlString])) {
+            $url->setId($urlString);
+
+            $this->urls[$urlString] = $url;
+            $this->pendingUrls[$urlString] = $url;
+        }
+
+        return $this;
+    }
+
+    public function hasPendingUrls() : bool
+    {
+        return (bool) $this->pendingUrls;
+    }
+
+    public function getUrlById($id) : CrawlUrl
+    {
+        if (! isset($this->urls[$id])) {
+            throw new UrlNotFoundByIndex("Crawl url $id not found in collection.");
+        }
+
+        return $this->urls[$id];
+    }
+
+    public function hasAlreadyBeenProcessed(CrawlUrl $url) : bool
+    {
+        $url = (string) $url->url;
+
+        return ! isset($this->pendingUrls[$url]) && isset($this->pendingUrls[$url]);
+    }
+
+    public function markAsProcessed(CrawlUrl $crawlUrl)
+    {
+        $url = (string) $crawlUrl->url;
+
+        unset($this->pendingUrls[$url]);
+    }
+
+    /**
+     * @param CrawlUrl|UriInterface $crawlUrl
+     *
+     * @return bool
+     */
+    public function has($crawlUrl) : bool
+    {
+        if ($crawlUrl instanceof CrawlUrl) {
+            $url = (string) $crawlUrl->url;
+        } else if ($crawlUrl instanceof UriInterface) {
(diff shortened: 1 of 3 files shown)


freekmurze:
After only having read the description of this PR, it already sounds very promising. 👍 I'll go through the code now.

freekmurze:
Pretty cool, thanks for your work on this. 🙌 

 I'll pull it in, polish it a bit and tag a new release.

src/CrawlQueue/ArrayCrawlQueue.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        return $this->urls[$id];
+    }
+
+    public function hasAlreadyBeenProcessed(CrawlUrl $url) : bool
+    {
+        $url = (string) $url->url;
+
+        return ! isset($this->pendingUrls[$url]) && isset($this->pendingUrls[$url]);
```
freekmurze: Maybe it's because I don't have had coffee yet, but I don't understand this line. Could you explain what you're checking here?

freekmurze: Added some tests around this and fixed this on `master`

src/CrawlQueue/ArrayCrawlQueue.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    protected $urls = [];
+
+    /**
+     * Pending URLs, indexed by URL string.
+     *
+     * @var CrawlUrl[]
+     */
+    protected $pendingUrls = [];
+
```
AbdelkaderBah: Add
```protected $processedUrls = []```

File: src/CrawlQueue/ArrayCrawlQueue.php (hunk 1 above)
Comment (AbdelkaderBah) [reply src/CrawlQueue/ArrayCrawlQueue.php]: - fix line 61: 
```return ! isset($this->pendingUrls[$url]) && isset($this->processedUrls[$url]);```

File: src/CrawlQueue/ArrayCrawlQueue.php [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $url = (string) $url->url;
+
+        return ! isset($this->pendingUrls[$url]) && isset($this->pendingUrls[$url]);
+    }
+
+    public function markAsProcessed(CrawlUrl $crawlUrl)
+    {
+        $url = (string) $crawlUrl->url;
+
```
AbdelkaderBah: - add ``$url`` to ``$this->processedUrls``

freekmurze: @AbdelkaderBah could you PR your proposals (with tests) to master?

src/CrawlQueue/ArrayCrawlQueue.php
BenMorel: Looks like you figured it out in 9078d286aac9b96f9ba4ed2fc66d82a3237c495f ! This line is the strict translation of:

https://github.com/spatie/crawler/blob/287d8525b613d7301e411f4a22df952aba64df63/src/CrawlQueue/CollectionCrawlQueue.php#L71

But less formatted for readability.

src/CrawlQueue/ArrayCrawlQueue.php
BenMorel: There is no real justification for an additional `$processedUrls` here: processed URLs are those that are in `$urls` and not in `$pendingUrls`, the check is easy and with a very low performance cost.

OTOH, adding an extra property to store these would **double the memory cost** of the crawl queue.

What could be done is *replacing* `$urls` with `$processedUrls`; so that pending URLs are only stored in `$pendingUrls`, and moved to `$processedUrls` once done. I'm not sure if this is a big improvement, but it can be done.

freekmurze: Ok, let's leave it like it is for now. Thanks for your feedback.

src/CrawlQueue/ArrayCrawlQueue.php
BenMorel: Thanks for adding some tests and for releasing, @freekmurze! 👍

Note that I'm not fond of adding an `InvalidUrl` exception: this is not an exception that's meant to be caught in userland, it can only happen if there is a bug in the code. For this reason, I think using built-in exceptions (`InvalidArgumentException`, or better yet in PHP 7, `TypeError` for consistency with PHP's built-in type checker) is the way to go.

If the justification is for moving the exception message logic out of the main code flow, I think a separate factory method for the exception would be appropriate.

Anyway it's released now, let's discuss this later in v5!

BenMorel:
Thanks for adding some tests and for releasing, @freekmurze! 👍

Note that I'm not fond of adding an `InvalidUrl` exception: this is not an exception that's meant to be caught in userland, it can only happen if there is a bug in the code. For this reason, I think using built-in exceptions (`InvalidArgumentException`, or better yet in PHP 7, `TypeError` for consistency with PHP's built-in type checker) is the way to go.

If the justification is for moving the exception message logic out of the main code flow, I think a separate factory method for the exception would be appropriate.

Anyway it's released now, let's discuss this later in v5!

freekmurze:
@BenMorel just wanted to let you know that already use the changes by this PR over at [Oh Dear!](https://ohdear.app). We see that our app now uses dramatically less CPU. Thanks to this change we can run more checks concurrently. 👍

BenMorel:
@freekmurze Wow, happy to hear that! 🎉

src/CrawlQueue/ArrayCrawlQueue.php
juancho48: isset($this->pendingUrls[$url]) is repeated twice? doesn't this return false always?

BenMorel: OMG I'm just seing it now, someone had to put it twice right under my eyes before I could see it. Thanks for pointing this out, it should have been `&& isset($this->urls[$url])`. This has been fixed in 9078d286aac9b96f9ba4ed2fc66d82a3237c495f anyway!

juancho48: I just wanna help out... cool thing is fixed

PR #254, opened 2019-08-07
Split $pendingUrls and $processedUrls

BenMorel:
Following @AbdelkaderBah's [comment](https://github.com/spatie/crawler/pull/250#discussion_r311432548), I wanted to give a try to a split of pending URLs and processed URLs.

It does simplify, for example, `hasAlreadyBeenProcessed()`, but adds some checks in other places.

All in all it's very similar, I'm not sure which implementation looks better. @freekmurze, your choice! ;-)

### DIFF
--- a/src/CrawlQueue/ArrayCrawlQueue.php
+++ b/src/CrawlQueue/ArrayCrawlQueue.php
@@ -10,27 +10,25 @@ use Spatie\Crawler\Exception\UrlNotFoundByIndex;
 class ArrayCrawlQueue implements CrawlQueue
 {
     /**
-     * All known URLs, indexed by URL string.
-     *
-     * @var CrawlUrl[]
-     */
-    protected $urls = [];
-
-    /**
      * Pending URLs, indexed by URL string.
      *
      * @var CrawlUrl[]
      */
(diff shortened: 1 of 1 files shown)


freekmurze:
I'll go for the new version. Thanks again! 👍

BenMorel:
Oh, by the way, this is a BC break I'm afraid (change to protected properties); this will require a 4.7.0 release.

BenMorel:
Actually I'm saying nonsense. This a BC break so would require a move to v5.

Maybe you should move this commit to v5-wip instead, and revert this in master?

BenMorel:
... and this conflicts with the work already done on `v5-wip`. Oops. Should we just revert this?

freekmurze:
Reverted, nice catch, didn't notice the BC break. Luckily I didn't tag it yet, so it's all good now.

BenMorel:
Thanks! This won't make a difference in v5 anyway, given the changes already been done.

By the way, do you have a strategy to merge changes made to `master` into `v5-wip` since it has been branched?

Some changes will need to be merged to v5, but some of them (like the present one) will somehow conflict.

PR #256, opened 2019-08-07
Fix inversion of logic & property name typo

BenMorel:
Just noticed 2 gross bugs that need a urgent fix:

- There was a logic inversion: it must return `true` if the URL is present in `$this->urls`
- `$this->urls` has been mispelt as `$this->url`, which went unnoticed due to `isset()`

I also updated the tests to catch this.

Could you please merge and tag a release ASAP? Thank you!


freekmurze:
Thank you!

BenMorel:
Thanks for your responsiveness! 👍

PR #262, opened 2019-09-18
Improvement: if a broken URL was found on multiple sources, list all sources

mattiasgeniar:
This fixes a situation like this:

```
http://localhost:9000/page1 - - +
                                |
                                |
                                + -- + --> http://localhost:9000/page3
                                |
                                |
http://localhost:9000/page2 - - +
```

2 pages both refer to a `/page3` URL. In the crawler results, it would only contain one instance for the `/page3` URL:

```
CrawledUrl {
  +crawledUrl: "http://localhost:9000/page3"
  +statusCode: 200
  +foundOnUrl: "http://localhost:9000/page1"
}
```

If we want to use the crawler to build a full graph of a website, linking every page/node to any other page/node, we need to store multiple `crawledUrl` & `foundOnUrl` combinations.

This change modifies the index of the found URLs arrays, by including the `foundOnUrl` value. This causes a page to be unique based on where it was found, thus showing up in a case like the one above. The crawler output for `/page3` would now be:

```
CrawledUrl {
  +crawledUrl: "http://localhost:9000/page3"
  +statusCode: 200
  +foundOnUrl: "http://localhost:9000/page1"
}

CrawledUrl {
  +crawledUrl: "http://localhost:9000/page3"
  +statusCode: 200
  +foundOnUrl: "http://localhost:9000/page2"
}
```

Showing every page that particular URL was found on.

Even though all previous tests pass, this is probably a breaking change. It will start to report more pages in the output.

### DIFF
--- a/src/CrawlQueue/ArrayCrawlQueue.php
+++ b/src/CrawlQueue/ArrayCrawlQueue.php
@@ -25,13 +25,13 @@ class ArrayCrawlQueue implements CrawlQueue
 
     public function add(CrawlUrl $url) : CrawlQueue
     {
-        $urlString = (string) $url->url;
+        $id = (string) $url->getId();
 
-        if (! isset($this->urls[$urlString])) {
-            $url->setId($urlString);
+        if (! isset($this->urls[$id])) {
+            $url->setId($id);
 
-            $this->urls[$urlString] = $url;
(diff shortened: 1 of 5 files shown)


mattiasgeniar:
Found a logical error, will fix & create new PR

PR #274, opened 2020-01-28
Allow Guzzle 7

alexeyshockov:
Guzzle 7 will have minimum BC breaks, and it looks like all the usages in the project are safe


alexeyshockov:
I'm not sure why Scrutinizer fails here...

freekmurze:
Thanks!

PR #280, opened 2020-02-21
WIP: add new observer method that is called when a link is found *again*

bobemoe:
I have found I needed to add this to allow my `CrawlObserver` to count the number of times a link is found, and also collate all the `foundOnUrl`'s for the given link.

I can now report *all* pages that link to a specific 404 rather than just the first one it was found on.

If this was already possible please let me know, I couldn't find another way :)

I can add tests and docs if you'd be interested in merging?

I guess adding new abstract function would break BC on all existing observer implementations, this would need to be major release?

### DIFF
--- a/src/CrawlObserver.php
+++ b/src/CrawlObserver.php
@@ -31,7 +31,7 @@ abstract class CrawlObserver
     );
 
     /**
-     * Called when the crawler has found the url again
+     * Called when the crawler has found the url again.
      *
      * @param \Psr\Http\Message\UriInterface $url
      * @param \Psr\Http\Message\ResponseInterface $response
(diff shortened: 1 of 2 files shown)


Redominus:
You could implement a new queue that counts how many times am Url has been found.

bobemoe:
I did consider that, I think I would have to use `CrawlQueue::has` as the logic that checks if it is already queued is in `Crawler::addToCrawlQueue`, the issue with this is that the `foundOnUrl` is not passed in, so although I could count, I wouldn't be able to collate the `foundOnUrl`'s.

bobemoe:
I am still using this as a patch in https://git.jhodges.co.uk/jhodges/sitemap and have still found no other way to achieve this otherwise.

bobemoe:
Not sure how I reopen this as suggested by spatie-bot?

PR #281, opened 2020-03-09
Bugfix: respect maximum response size when checking Robots Meta tags

mattiasgeniar:
Large HTML bodies would be passed as-is to the `RobotsMeta` check. If there was a 100MB HTML blob, it would be passed as a giant string to that class, causing out of memory issues further down the stack.

This (breaking) change modifies this behaviour, by respecting the maximum responze size (default: 2MB) when checking the body for HTML meta tags.

Alternate implementation would be to extract `convertBodyToString()` to a helper function, so it could be called from both `CrawlRequestFulfilled` and `CrawlerRobots`. However, it looks like `CrawlerRobots` is only used internally here, the breaking change might not actually affect anyone.

### DIFF
--- a/src/CrawlerRobots.php
+++ b/src/CrawlerRobots.php
@@ -17,11 +17,11 @@ class CrawlerRobots
     /** @var bool */
     protected $mustRespectRobots;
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #285, opened 2020-03-21
Added option to customize xpath

NikosDevPhp:
This is DPR making it able to customize the xpath value for a given Crawl.
This is just a draft, but I think it can come in handy for a lot of use cases.
e.g. Distinguishing product pages from headert,footer,sociallinks etc in ecommerce crawling

squashed commits: 
created option to customize xpath value, 
added tests for xpath to assert only crawls specific xpath, 
updated README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,16 @@ This package comes with three `CrawlProfiles` out of the box:
 - `CrawlInternalUrls`: this profile will only crawl the internal urls on the pages of a host.
 - `CrawlSubdomains`: this profile will only crawl the internal urls and its subdomains on the pages of a host.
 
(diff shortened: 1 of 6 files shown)


NikosDevPhp:
It would be great if you could take a look and code review as I am not very experienced or prompt me to some direction, but I think this functionality would be of great use since it allows average crawling urls to significantly drop which can improve performance and also less requests means less bans.

CrawlProfile class should be able to distinguish pages(maybe according to depth) to provide different xpaths. This is a basic use case for example for Ecommerce crawling.
Could work more on that.

If you could provide me some feedback it would be great

Issue #283, opened 2020-03-18
The "shouldCrawl" method of the "CrawlProfile" interface is executed more than one time.


design-principles:
I've installed version 4.6 of the crawler and ran into one problem.
The `shouldCrawl` method is executed twice.
`shouldCrawl` is executed twice when I pass 1 to the `setMaximumCrawlCount` method. When I pass 2, `shouldCrawl` is executed 4 times and so on.
Why this is happening? Is that a bug or a feature?

brendt:
@design-principles It's called two times indeed:

- In `Crawler` > `startCrawling`  (line 386) > `addToCrawlQueue` (line 481)
- In `Crawler` > `startCrawling` (line 391) > `startCrawlingQueue` (line 430) > `getCrawlRequests` (line 460)

I'm not sure why the second check is done. @freekmurze you wrote it in back in 2016, but the commit messages says `very wip` 😂

@design-principles can you elaborate why you opened this issue? Are you running into any problems because of this?

freekmurze:
lol #wip

design-principles:
@brendt I've got some complex logic that is executed inside `shouldCrawl`. And yes, my script loses performance because of it.

brendt:
@freekmurze do you think it's worth the risk removing the second check? All tests seem to work, though I'm a little hesitant because this is a very old piece of the codebase. A more sage approach would be to incorporate this change in a new major version, but I don't think we had any concrete plans for a major release soon.

freekmurze:
Can you send a PR with this change. I'll test it out in Oh Dear. If everything works there, I'll merge the change and tag a new release.

brendt:
@freekmurze here it is: https://github.com/spatie/crawler/pull/289

freekmurze:
I think we need the existing logic to cover some edge cases.

@design-principles if you're doing some expensive work inside that `shouldCrawl` method, I suggest you manually cache the results somewhere.

brendt:
@design-principles please refer to https://github.com/spatie/crawler/pull/289, we're going to leave the check in. 

To solve your problem, I suggest to cache the expensive logic, could be as simple as a static array, with a mapping of all URLs and whether they should be crawled or not.

PR #289, opened 2020-03-25
Remove double should crawl check

brendt:
Possible fix for #283

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -457,11 +457,6 @@ class Crawler
     protected function getCrawlRequests(): Generator
     {
         while ($crawlUrl = $this->crawlQueue->getFirstPendingUrl()) {
-            if (! $this->crawlProfile->shouldCrawl($crawlUrl->url)) {
-                $this->crawlQueue->markAsProcessed($crawlUrl);
-                continue;
-            }
-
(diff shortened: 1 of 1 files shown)


brendt:
Hang on, turns out some tests do fail.

brendt:
These were failing before this PR though: https://travis-ci.org/github/spatie/crawler/jobs/660322726?utm_medium=notification&utm_source=github_status

src/Crawler.php
```diff
@@ -457,11 +457,6 @@ protected function createRobotsTxt(UriInterface $uri): RobotsTxt
     protected function getCrawlRequests(): Generator
     {
         while ($crawlUrl = $this->crawlQueue->getFirstPendingUrl()) {
-            if (! $this->crawlProfile->shouldCrawl($crawlUrl->url)) {
-                $this->crawlQueue->markAsProcessed($crawlUrl);
-                continue;
```
freekmurze: Shouldn't we still have this `continue` though? Because with this change all logic in this while loop will get executed if the profile determines that it shouldn't crawl the url.

freekmurze:
I think I explicitly added this check to take care of some edge cases. I'm going to take the safe route and leave it in.

PR #292, opened 2020-04-10
Fix LinkAdder not receiving the updated DOM

chinleung:
I'm using [spatie/laravel-sitemap](https://github.com/spatie/laravel-sitemap) for a SPA and noticed that the sitemap is no longer generated properly. It's always a single node xml with the homepage only.

After digging, I've found that the LinkAdder instance never receives the DOM after executing the JavaScript, therefore the other urls added after executing JavaScript are never crawled.

### DIFF
--- a/src/Handlers/CrawlRequestFulfilled.php
+++ b/src/Handlers/CrawlRequestFulfilled.php
@@ -42,9 +42,9 @@ class CrawlRequestFulfilled
         $crawlUrl = $this->crawler->getCrawlQueue()->getUrlById($index);
 
         if ($this->crawler->mayExecuteJavaScript()) {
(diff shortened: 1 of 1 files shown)


freekmurze:
I don't see how renaming this variable fixes that potential problem.

chinleung:
> I don't see how renaming this variable fixes that potential problem.

@freekmurze Because at line 66 there's this:

```php
$this->linkAdder->addFromHtml($body, $baseUrl);
```

And since the variable is `$html` instead of `$body`, it's passing the DOM before the JavaScript has been rendered instead of the new one.

freekmurze:
Gotcha, thanks for explaining. I didn't notice the body being used in the rest of the function (was collapsed on GitHub).

PR #293, opened 2020-04-12
Add support for setParseableMimeTypes()

mattiasgeniar:
By default, every found page will be downloaded (up to `setMaximumResponseSize()` in size) and parsed for additional links. You can limit which content-types should be downloaded and parsed by setting the `setParseableMimeTypes()` with an array of allowed types.

```php
Crawler::create()
    ->setParseableMimeTypes(['text/html', 'text/plain'])
```

This will prevent downloading the body of pages that have different mime types, like binary files, audio/video, ... that are unlikely to have links embedded in them. This feature mostly saves bandwidth.

### DIFF
--- a/tests/CrawlerTest.php
+++ b/tests/CrawlerTest.php
@@ -437,4 +437,33 @@ class CrawlerTest extends TestCase
 
         $this->assertSame($newUserAgent, $actualUserAgent);
     }
+
+    /** @test */
+    public function it_will_only_crawl_correct_mime_types_when_asked_to()
+    {
+        Crawler::create()
+            ->setCrawlObserver(new CrawlLogger())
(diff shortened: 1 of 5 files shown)


tests/CrawlerTest.php
```diff
@@ -437,4 +437,33 @@ public function it_should_ignore_user_agents_header_case()
 
         $this->assertSame($newUserAgent, $actualUserAgent);
     }
+
+    /** @test */
+    public function it_will_only_crawl_correct_mime_types_when_asked_to()
```
freekmurze, line 442: Very nice 👍

freekmurze:
Perfect PR, thanks!

PR #295, opened 2020-04-14
Allow curl streaming responses

mattiasgeniar:
A simple change, but necessary to allow the `stream => true` option to stream responses instead of downloading upfront.


freekmurze:
Thanks!

PR #297, opened 2020-04-15
Improvement: pass the body to each crawled() implementation

mattiasgeniar:
Problem: if you use a Crawler Client with Curls' streaming implementation, the body can only be read once.

Ie:

```
$clientConfig = ['stream' => true];

Crawler::create($clientConfig)
            ->setCrawlObserver(new CrawlLogger())
            ->startCrawling('https://site.tld');
```

Each crawled page will have a `$response` object with a `getBody()` method. But since it's using PHP's streams, the body can only be read once.

If you have multiple observers, only the first observer will have access to the body.

This PR adds the ability to pass the body to each observer, ensuring every observer has access to the full body as limited by `setMaximumResponseSize()` and `setParseableMimeTypes()`.

Downside: this is a breaking change, as it changes the signatures of the `crawled()` implementations.

### DIFF
--- a/src/CrawlObserver.php
+++ b/src/CrawlObserver.php
@@ -27,7 +27,8 @@ abstract class CrawlObserver
     abstract public function crawled(
         UriInterface $url,
         ResponseInterface $response,
-        ?UriInterface $foundOnUrl = null
+        ?UriInterface $foundOnUrl = null,
+        string $body = ""
     );
 
     /**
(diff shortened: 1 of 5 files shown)


freekmurze:
@brendt do you see a way of making this non-breaking. 

I think we want this change, but will have to create a new major version. There's seemingly no way around that.

freekmurze:
I'll merge this in a `v5` branch, so we can do some more breaking changes.

Redominus:
Sorry for being too late to the party. 
There is something I do not understand about this feature. 
By allowing the usage of streams, now we are copying a very long string as many times as observers we have. The same job a seekable stream does just moving an object reference around.
Does allowing streams provide any benefit to the crawling process?

PR #300, opened 2020-05-05
Link Adder; only add links that can be completely parsed

mattiasgeniar:
Situation: now that there is an ability to only parse the first X bytes of a page, it's possible that there is a link that gets parsed that isn't entirely valid.

Take this configuration:

```
Crawler::create()
    ->setMaximumResponseSize(1024 * 512)
```

This will parse the first 512KB of a page. This means the body that the crawler is working with, might contain something like this:

```
<HTML>
...
<ul>
  <li> <a href="/link1">link #1</a> </li>
  <li> <a href="/link2">link #2</a> </li>
  <li> <a href="/li
```

That third item should obviously point to `/link3`, but since we cut the dom at 512KB, it might not be complete there.

This PR prevents the parsing of that last link, by checking if there is;
1. A HTML element _after_ the link that is being parsed
2. The found HTML element contains at least 1 child node (the text between <a></a>)

The Symfony DOMCrawler accepts non-fully closed `href` tags as valid, which causes weird situations if the HTML DOM is only partially parsed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ By default, every found page will be downloaded (up to `setMaximumResponseSize()
 
 ```php
 Crawler::create()
-    ->setParseableMimeTypes(['text/html', 'text/plain']) 
+    ->setParseableMimeTypes(['text/html', 'text/plain'])
 ```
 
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks!

PR #303, opened 2020-05-25
BodyStream: read the data in chunks

mattiasgeniar:
Bug: not every stream allows all requested data for the page body to be read in a single go.

This PR resolves this by loading data in chunks of 512 bytes, until the maximum of `$readMaximumBytes` is reached.

Functionally, this doesn't change anything to the parser, it just loads the HTML body more efficiently.

I failed to come up with a good test for this in nodejs, as it appears to be a very low-level implementation issue that causes streams to only accept small chunks of data. The error occurs on data hosted at Google's Storage API's.


freekmurze:
Thanks!

PR #305, opened 2020-06-08
Bugfix: import the correct Exception namespace

mattiasgeniar:
Subtle bug: the `Exception` wasn't being caught as it wasn't imported.


freekmurze:
Thanks!

PR #307, opened 2020-06-24
Enhancement: Update nicmart/tree

localheinz:
This PR

* [x] updates [`nicmart/tree`](https://github.com/Nicmart/tree)

💁‍♂️ I am currently modernizing `nicmart/tree`, and I assume this project here is one of the largest consumers. Would be good to see if modernizing it makes sense (because it is actually used).


freekmurze:
Thanks!

localheinz:
Thank you, @freekmurze!

PR #308, opened 2020-06-24
Fix: Remove reference

localheinz:
This PR

* [x] removes a reference to an environment variable that is not set


freekmurze:
Thanks!

localheinz:
Thank you, @freekmurze!

PR #309, opened 2020-06-24
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] normalizes `composer.json`

💁‍♂️ I ran

```
$ composer global require ergebnis/composer-normalize
```

followed by

```
$ composer normalize
```

For reference, see https://github.com/ergebnis/composer-normalize.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,11 +32,6 @@
     "config": {
         "sort-packages": true
     },
-    "extra": {
-        "branch-alias": {
-            "dev-master": "0.0.x-dev"
-        }
-    },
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

localheinz:
Thank you, @freekmurze!

In case there is interest, I will be happy to add it to the pipeline!

freekmurze:
Not necessary, I don't think the extra weight would be worth the benefits. Thanks for suggesting it though.

PR #310, opened 2020-07-15
Add crawler option to allow crawl links with rel="nofollow"

barocode:
Sometimes it may be necessary to get links containing the attribute rel="nofollow". 
I think it should be possible to allow the collection of such links.


freekmurze:
Thanks!

PR #321, opened 2020-10-05
Fix incorrect CrawlProfile namespace in readme

kmcluckie:
Just noticed this small issue when setting up the crawler.


freekmurze:
Thanks!

PR #329, opened 2020-10-21
Readme: Fix wrong namespace for CrawlObserver

bjoernhasse:
I noticed that `\Spatie\Crawler\CrawlObserver` is not a correct namespace.


freekmurze:
Thank you!

PR #343, opened 2020-12-19
Fix support for PHP 8.0 in v4

Bapawe:
In order to fix https://github.com/spatie/http-status-check/issues/77 i ran into several issues in this package which i fixed in this PR for v4.

I still have to fix v5 and v6 since they lack the following in order to actually test the code with prefer-lowest dependencies:
- composer: update minimum guzzle and dom-crawler version in order to support php 8
- run-tests: Add missing dependency-version to the composer step

I'll try to create PR's for them tomorrow. 

Plz let me know if i fixed this correctly or i should have done something differently ;)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -9,7 +9,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [ubuntu-latest]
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you!

PR #345, opened 2020-12-19
Fix support for PHP 8.0 in v5

Bapawe:
- composer: update minimum guzzle and dom-crawler version in order to support php 8
- run-tests: Add missing dependency-version to the composer step


freekmurze:
Thanks!

PR #364, opened 2021-04-26
Add PHP 8-only Support (v7)

patinthehat:
This PR adds a new major version, v7.0.0.

Specifically, it:

- Removes support for all PHP 7.x versions.
- Requires PHP 8.0+.
- Uses PHP 8 syntax where possible.
- Adds missing typehints/return types.
- Removes unnecessary PHP docblocks per Spatie's guidelines.
- Adds/updates several files to more closely match `spatie/package-skeleton-php`.
- Bumps the minimum `PHPUnit` version to `9.5`.
- Bumps the version of `spatie/robots-txt` to `2.0`, which is a PHP 8-only version.
- Fixes minor issues in the readme.
- _Updates the changelog - release date needs to be updated, currently is "unreleased"._


freekmurze:
Thank you very much again!

PR #368, opened 2021-05-10
Added laravel application as exemple of using laravel model as queue

insign:
The main reason is the others queues packages store all items in one single array, which can be a RAM problem for big sites. Furthermore, you can preserve and use crawled links as you want.


freekmurze:
Thanks!

Issue #379, opened 2021-10-18
[v7.0.2] Undefined method stream_for


flangofas:
Should be replaced with `\GuzzleHttp\Psr7\Utils::streamFor` 

Used by [CrawlRequestFulfilled:#43](https://github.com/spatie/crawler/blob/main/src/Handlers/CrawlRequestFulfilled.php#L43)

PR #391, opened 2021-11-12
Replace stream_for with Utils::streamFor

NicolasMica:
Hi 👋 I just encountered the same error as #379 so I figured I would submit a PR to fix it as it was quite simple.
This supports the GuzzleHttp [function API upgrade](https://github.com/guzzle/psr7#upgrading-from-function-api) to 2.0.


freekmurze:
Thank you!

PR #392, opened 2021-11-15
Keep only guzzlehttp/psr7 v2.0

flangofas:
Relates to https://github.com/spatie/crawler/issues/379

Bump `guzzlehttp/psr7` to `v2.0` after the change in https://github.com/spatie/crawler/pull/391

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": "^8.0",
         "guzzlehttp/guzzle": "^7.3",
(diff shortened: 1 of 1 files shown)


freekmurze:
The current testsuite seems to run fine for the current dependencies, so I'm not sure why this change is needed.

flangofas:
```Utils::streamFor``` is only available in `guzzlehttp/psr7` v2.0 therefore, this change

freekmurze:
Thanks!

Issue #395, opened 2022-03-19
Deprecations of implemented interfaces since PHP 8.1 for `CrawlObserverCollection` class


BrokenSourceCode:
Hi,

This issue is related to [this one](https://github.com/spatie/laravel-sitemap/issues/440) from the `spatie/laravel-sitemap` repository. Since the **PHP 8.1** version, the [`CrawlObserverCollection` class](https://github.com/spatie/crawler/blob/main/src/CrawlObservers/CrawlObserverCollection.php) emits deprecations due to the implemented interfaces.

```log
PHP Deprecated:  Return type of Spatie\Crawler\CrawlObservers\CrawlObserverCollection::offsetExists($offset) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/vendor/spatie/crawler/src/CrawlObservers/CrawlObserverCollection.php on line 66
PHP Deprecated:  Return type of Spatie\Crawler\CrawlObservers\CrawlObserverCollection::offsetGet($offset) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/vendor/spatie/crawler/src/CrawlObservers/CrawlObserverCollection.php on line 52
PHP Deprecated:  Return type of Spatie\Crawler\CrawlObservers\CrawlObserverCollection::offsetSet($offset, $value) should either be compatible with ArrayAccess::offsetSet(mixed $offset, mixed $value): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/vendor/spatie/crawler/src/CrawlObservers/CrawlObserverCollection.php on line 57
PHP Deprecated:  Return type of Spatie\Crawler\CrawlObservers\CrawlObserverCollection::offsetUnset($offset) should either be compatible with ArrayAccess::offsetUnset(mixed $offset): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/vendor/spatie/crawler/src/CrawlObservers/CrawlObserverCollection.php on line 71
[...]

PR #396, opened 2022-03-19
Fix issue #395

BrokenSourceCode:
Hi,

This short pull request fix the issue [https://github.com/spatie/crawler/issues/395](https://github.com/spatie/crawler/issues/395).

Methods signatures from the PHP documentation:
- [`ArrayAccess`](https://www.php.net/manual/en/class.arrayaccess.php#arrayaccess.synopsis)
```php
interface ArrayAccess {
	/* Methods */
	public offsetExists(mixed $offset): bool
	public offsetGet(mixed $offset): mixed
	public offsetSet(mixed $offset, mixed $value): void
	public offsetUnset(mixed $offset): void
}
```
- [`Iterator`](https://www.php.net/manual/en/class.iterator.php#iterator.synopsis)
```php
interface Iterator extends Traversable {
	/* Methods */
	public current(): mixed
	public key(): mixed
	public next(): void
	public rewind(): void
	public valid(): bool
}
```

Regards.

### DIFF
--- a/src/CrawlObservers/CrawlObserverCollection.php
+++ b/src/CrawlObservers/CrawlObserverCollection.php
@@ -44,17 +44,17 @@ class CrawlObserverCollection implements ArrayAccess, Iterator
         }
     }
 
-    public function current()
+    public function current(): mixed
     {
         return $this->observers[$this->position];
     }
 
-    public function offsetGet($offset)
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

Issue #352, opened 2021-02-12
setParseableMimeTypes() not working


flmommens:
Hi,
I'm using spatie/crawler 4.7.5 with php 7.2

I'm trying to exclude non-html documents from the crawl with

`Crawler::create()->setParseableMimeTypes(['text/html'])
`

The documentation says this method "will prevent downloading the body of pages that have different mime types".
However, the crawl method of my Observer gets called with URL of images, so I assume images are still being crawled.

Is there something I haven't understood?
Thanks

freekmurze:
This test proves that the functionality is working correctly
https://github.com/spatie/crawler/blob/17beb464169487304007f6d856f6ef1e42800ad7/tests/CrawlerTest.php#L478

Could you submit a failing test for us to look at?

flmommens:
```
<?php
require __DIR__ . '/vendor/autoload.php';
use Spatie\Crawler\Crawler;
use Spatie\Crawler\CrawlInternalUrls;
use Spatie\Crawler\CrawlObserver;
use GuzzleHttp\Exception\RequestException;
use GuzzleHttp\Psr7;
use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\UriInterface;

class IndexerCrawlObserver extends CrawlObserver
{
    protected $indexer;
    protected $indexName;
    
    public function crawled(UriInterface $url, ResponseInterface $response, ?UriInterface $foundOnUrl = null)
    {
        echo "Crawled $url\n";
    }

    public function crawlFailed(UriInterface $url, RequestException $e, ?UriInterface $foundOnUrl = null)
    {
        echo sprintf("Failed crawling %s %s\n", $url, Psr7\Message::toString($e->getResponse()));
    }
}

class SpatieCrawler
{
    protected $crawler;
    
    public function __construct(ContainerBagInterface $bag = null)
    {
        $this->crawler = Crawler::create()->acceptNofollowLinks()->setParseableMimeTypes(['text/html']);
    }

    public function setCrawlObserver(CrawlObserver $observer)
    {
        $this->crawler->setCrawlObserver($observer);
    }
    
    public function crawl($url)
    {
        $this->crawler->setCrawlProfile(new CrawlInternalUrls($url));
        $this->crawler->startCrawling($url);
    }
}

$crawler = new SpatieCrawler();
$crawler->setCrawlObserver(new IndexerCrawlObserver());
$crawler->crawl('https://marketmegood.com/blog/');
```
[...]

PR #432, opened 2022-10-19
Make setParseableMimeTypes prevent URLs from being crawled

arturzealves:
### Some context
I started using this package yesterday and got the same problem that was also reported on issue https://github.com/spatie/crawler/issues/352

This is probably not the best way to fix the issue, but I don't have enough knowledge and time to do it properly right now.
So if this pull request is not good enough, point me in the right direction or feel free to continue this work, so that we can have a mergeable pull request sooner.

I didn't add new tests because I wasn't able to have the test server running.

### Description
What I did was to check if the response of `getBody` was an empty string and do an early return if it is.
The main purpose was to avoid the execution of [line#41](https://github.com/arturzealves/crawler/blob/main/src/Handlers/CrawlRequestFulfilled.php#L41) and [line#50](https://github.com/arturzealves/crawler/blob/main/src/Handlers/CrawlRequestFulfilled.php#L50).

I was a bit afraid that maybe the `ArrayCrawlQueue` may start to have pending URLs, because [line#38](https://github.com/arturzealves/crawler/blob/main/src/Handlers/CrawlRequestFulfilled.php#L38) is not being executed anymore, but I did some debug while testing manually and with this change, there are no pending URLs in the end.

### How to reproduce
The code I had running on my computer that made me find this issue comes from [this page](https://anlisha.com.np/blog/web-scraping-with-laravel-and-spatie-crawler/). You will have to uncomment the line related with the `setParseableMimeTypes` of course.
On that code example, it will crawl on `https://www.lipsum.com` and with this fix, it won't crawl URLs of `.pdf` or `.gz` files.
 
Screenshot 2022-10-19 at 23 20 30

As seen above, in my example, there are no log messages saying `crawled` of URLs that point to files

### DIFF
--- a/src/Handlers/CrawlRequestFulfilled.php
+++ b/src/Handlers/CrawlRequestFulfilled.php
@@ -28,6 +28,9 @@ class CrawlRequestFulfilled
     public function __invoke(ResponseInterface $response, $index)
     {
         $body = $this->getBody($response);
+        if (empty($body)) {
+            return;
+        }
 
         $robots = new CrawlerRobots(
             $response->getHeaders(),


freekmurze:
Thanks for your work on this. I'm considering merging this in, but the tests need to pass.

Could you rebase with the `main` branch to allow those composer plugins to run?

XAKEPEHOK:
Why this was not merged yet?

Issue #259, opened 2019-08-26
Is it possible to modify the URL to be crawled?


mw108:
Is it possible to modify the URL, which will be crawled, for instance, add "index.php" at the end of the URL?

URL to be crawled: https://test.mysite.com/my-uri/
Modified URL that should be crawled instead: https://test.mysite.com/my-uri/index.php

freekmurze:
Though I've never used it that way, I think you can modify the url that's being passed to `willCrawl` of a crawl observer. Check the docs to learn how to create a a custom crawl observer.

codealfa:
Could we look back at this? What you are passed is an instance of `Psr\Http\Message\UrInterace`, which is immutable and `willCrawl` doesn't return any values so there's no easy way to modify the URL. This feature would be useful. Thanks.

PR #440, opened 2023-01-17
Modify URLs before crawl

codealfa:
**Overview**

As requested in issue #259, I also needed to modify the URLs to be crawled, but the solution suggested didn't work. The URL passed to `willCrawl` of an observer is an instance of `Psr\Http\Message\UriInterface`, which is immutable. Further, this method doesn't return any values, so there's no easy way to modify the URL. Even if you could modify the URL here, it doesn't work because you're now changing the id of the URL in the queue, so the crawler can no longer recognize if the URL was already crawled. The same URL then gets crawled over and over until the crawl limit is reached. We'd need to modify the URL before it is added to the crawl queue.

**Use Case**

I use the crawler in a plugin I develop to optimize websites, which modifies the HTML of the site on the run. Sometimes, I need the original HTML before modification by the plugin. I can achieve this by passing a query to the URL, such as 'nooptimize=1', which notifies the plugin that it shouldn't modify the HTML on this request. So I need to pass this query to each URL to be crawled so I can access the 'unoptimized' HTML while the plugin is optimizing the HTML for other requests.

**Solution**

I implemented this by adding a new method, `filterCrawlUrl`, to the abstract `CrawlObserver` that custom child observers can overwrite to make modifications if required. Otherwise, it simply passes the same URL back to the crawler. A new method, `Crawler::createCrawlUrl`, is assigned the responsibility to create the `CrawlUrl` and passes the created instance to each observer to filter. All instances of `CrawlUrl::create` are replaced by `Crawler::createCrawlUrl`. This way, the URL gets modified before it is added to the crawl queue.

**Caveat**

Since the links extracted from the HTML are added to the depth tree before they can be modified, you can't set a maximum depth when using the crawler in this way. Setting a maximum depth to any value effectively sets it to 1. You'd have to use some other means to limit the number of links crawled, such as `setCurrentCrawlLimit`, or `setTotalCrawlLimit`.

**Conclusion**

With this implementation, the crawler can now allow URLs to be modified before crawled. This also doesn't affect existing codes, as custom observers do not need to implement the `filterCrawlUrl` if not required. I've added a unit test which shows links being modified as expected and with existing tests passing without any modifications. 

I know it may seem a bit much as the crawler is very popular, and this doesn't seem like a much-requested feature, but it was essential for my use case, so I'm hoping it can be implemented. I'm currently using a modified fork of a previous version because I need to support PHP 7.4. Still, hopefully, in time, I can switch back to the main repo to benefit from added features and improvements.

### DIFF
--- a/src/CrawlObservers/CrawlObserver.php
+++ b/src/CrawlObservers/CrawlObserver.php
@@ -15,7 +15,7 @@ abstract class CrawlObserver
      *
      * @return \Psr\Http\Message\UriInterface
      */
-    public function filterCrawlUrl(UriInterface $url): UriInterface
+    public function modifyCrawlUrl(UriInterface $url): UriInterface
     {
         return $url;
     }
--- a/src/CrawlObservers/CrawlObserver.php
+++ b/src/CrawlObservers/CrawlObserver.php
@@ -9,6 +9,18 @@ use Psr\Http\Message\UriInterface;
 abstract class CrawlObserver
 {
     /**
+     * Called when CrawlUrl is created to allow for modification
+     *
+     * @param   \Psr\Http\Message\UriInterface  $url
+     *
+     * @return \Psr\Http\Message\UriInterface
+     */
+    public function filterCrawlUrl(UriInterface $url): UriInterface
+    {
+        return $url;
+    }
+
+    /**
      * Called when the crawler will crawl the url.
      *
      * @param \Psr\Http\Message\UriInterface $url
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -409,9 +409,9 @@ class Crawler
 
         $this->totalUrlCount = $this->crawlQueue->getProcessedUrlCount();
 
-        $this->baseUrl = $baseUrl;
+        $crawlUrl = $this->createCrawlUrl($baseUrl);
 
-        $crawlUrl = CrawlUrl::create($this->baseUrl);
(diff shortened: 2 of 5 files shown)


src/CrawlObservers/CrawlObserver.php
```diff
@@ -8,6 +8,18 @@
...
 abstract class CrawlObserver
 {
+    /**
+     * Called when CrawlUrl is created to allow for modification
+     *
+     * @param   \Psr\Http\Message\UriInterface  $url
+     *
+     * @return \Psr\Http\Message\UriInterface
+     */
+    public function filterCrawlUrl(UriInterface $url): UriInterface
```
freekmurze, line 18: Since the primary use case is modify the URL, let's rename this to `modifyCrawlUrl`.

freekmurze (changes requested):

freekmurze:
As adding a method to an abstract class is a breaking change, I'm not going to merge this in right now.

I'll revisit this when creating a new major version of the crawler. 

In the meantime you can keep using the code of your PR.

codealfa:
Understood. Thanks for reviewing, and I look forward to this feature being added sometime in the future.

Redominus:
In the meantime may I suggest using pushing an url rewrite middleware to the crawler's Guzzlehttp configuration?

freekmurze:
Feel free to PR that, make sure to don't make breaking changes, update the readme and update tests.

PR #449, opened 2023-11-18
Define only needed methods in observer implementation

buismaarten:
Hey there,

In some use cases the observer methods like `crawled` and `crawlFailed` are not always needed in the implementation.
This change allow the developer to only use the necessary methods without having to define other abstract methods.


freekmurze:
Thanks!

Issue #450, opened 2023-11-18
Crawler stops after `www.` instead of no subdomain


buismaarten:
While crawling a page, a link starting with 'www' is found. While the website redirects from 'www' to non-'www', the crawler simply stops. This is caused by the code below.

```php
if (! $this->crawler->getCrawlProfile() instanceof CrawlSubdomains) {
    if ($crawlUrl->url->getHost() !== $this->crawler->getBaseUrl()->getHost()) {
        return;
    }
}
```

The second if-statement evaluates to something like `if ('example.com' !== 'www.example.com')`, this causes the crawler to stop.

Note: the Guzzle option `allow_redirects` must be set to `true` to follow redirects.

PR #451, opened 2023-11-19
Fix for crawler stopping issue

buismaarten:
When the `CrawlSubdomains` profile is not used, no additional code is used to match some hostnames in my opinion. Is there a specific reason to include this code?

Fixes issue #450

### DIFF
--- a/src/Handlers/CrawlRequestFulfilled.php
+++ b/src/Handlers/CrawlRequestFulfilled.php
@@ -50,12 +50,6 @@ class CrawlRequestFulfilled
             $this->handleCrawled($responseWithCachedBody, $crawlUrl);
         }
 
-        if (! $this->crawler->getCrawlProfile() instanceof CrawlSubdomains) {
-            if ($crawlUrl->url->getHost() !== $this->crawler->getBaseUrl()->getHost()) {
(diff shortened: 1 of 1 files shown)


freekmurze:
I think the current code is necessary: in normal circumstances we want to stop crawling when we're not on our own domain anymore, but if `CrawlSubdomains` is used, it is allowed.

buismaarten:
When the `CrawlAllUrls` profile is used, is it needed to skip the URL? This crawler normally works when we're not on our own domain.

freekmurze:
Yes, it is needed. Otherwise the crawl would start crawling other domains and never stop.

buismaarten:
Maybe I don't fully understand it, but isn't the reason why we can specify a depth or a limit while crawling?

freekmurze:
No, the idea of the Crawler is that you only crawl one domain, and not go outside that domain by default.

buismaarten:
Is it an idea to make the if-statement configurable? Because the current crawler profiles cannot fix this problem at this moment.

freekmurze:
Feel free to send a PR with a concrete proposal + tests on how to handle this better

PR #452, opened 2023-11-20
Fix return type

riesjart:
`Spatie\Crawler\Crawler::startCrawling()` accepts any instance of `Psr\Http\Message\UriInterface`. 
When e.g. an instance of `Spatie\Url\Url` has been provided, everything goes well until `Spatie\Crawler\Handlers\CrawlRequestFulfilled::getBaseUrl()` fails to return an instance of `GuzzleHttp\Psr7\Uri` instead of the PSR interface.  The code doesn't depend on the concrete Guzzle `Uri` however.


freekmurze:
Thanks!

PR #460, opened 2024-02-06
Fix wording in documentation

adamtomat:
The sentence didn't make sense and had some words missing. This MR just adds those words in.


freekmurze:
Thanks!

PR #462, opened 2024-02-25
Add request factory

asbiin:
This adds the ability to define a request factory to create a Guzzle request


Redominus:
Hi @asbiin and thank you for your contribution.
the test shows a behavior that could be already achieved by creating the crawler like this
```PHP
Crawler::create([
   RequestOptions::HEADERS => [ 
         'X-Foo' => 'Bar'.
   ],
]);
```
Maybe a bit more code inside the factory would show the true potential of the feature

PR #467, opened 2024-07-01
Check original URL against depth tree when visited link is a redirect

### LINKED ISSUE #466: Using a maximum depth prevents redirected links from being crawled
If a maximum depth is specified while redirects are enabled and being tracked, any crawled links under a redirect will not be visited, since each link's parent URL will not match a node in the tree and will be discarded. This applies for cases where the initial crawl URL redirects, and where subsequent crawled URLs are redirects.

After a redirect gets followed, `CrawlRequestFulfilled` changes the base URL to the final link in the redirect history (presumably so that `$foundOnUrl` is most correct).
```php
protected function getBaseUrl(ResponseInterface $response, CrawlUrl $crawlUrl): UriInterface
{
    $redirectHistory = $response->getHeader(RedirectMiddleware::HISTORY_HEADER);

    if (empty($redirectHistory)) {
        return $crawlUrl->url;
    }

    return new Uri(end($redirectHistory));
}

// implementation from __invoke:
$baseUrl = $this->getBaseUrl($response, $crawlUrl);
$this->urlParser->addFromHtml($body, $baseUrl);
```
This is then passed to `LinkUrlParser`, which adds each URL to the depth tree.
```php
public function addFromHtml(string $html, UriInterface $foundOnUrl): void
{
    $allLinks = $this->extractLinksFromHtml($html, $foundOnUrl);

    collect($allLinks)
        ->filter(fn (Url $url) => $this->hasCrawlableScheme($url))
        ->map(fn (Url $url) => $this->normalizeUrl($url))
        ->filter(function (Url $url) use ($foundOnUrl) {
            // will always fail if redirect
            if (! $node = $this->crawler->addToDepthTree($url, $foundOnUrl)) {
                return false;
            }

            return $this->shouldCrawl($node);
        })
        ->filter(fn (Url $url) => ! str_contains($url->getPath(), '/tel:'))
            $crawlUrl = CrawlUrl::create($url, $foundOnUrl, linkText: $url->linkText());

            $this->crawler->addToCrawlQueue($crawlUrl);
        });
}
```
The parent (or base) node on the tree will still be the original URL that was found on the page or supplied by the user as the initial crawl URL. Since the parent URL differs from the node on the depth tree, no children will ever be added.
```php
public function addToDepthTree(UriInterface $url, UriInterface $parentUrl, ?Node $node = null): ?Node
{
    if (is_null($this->maximumDepth)) {
        return new Node((string) $url);
    }

    $node = $node ?? $this->depthTree;

    $returnNode = null;

    // this will never be true if the link was a redirect since $parentUrl won't be on the tree
    if ($node->getValue() === (string) $parentUrl) {
        $newNode = new Node((string) $url);

        $node->addChild($newNode);

        return $newNode;
    }

    foreach ($node->getChildren() as $currentNode) {
        $returnNode = $this->addToDepthTree($url, $parentUrl, $currentNode);

        if (! is_null($returnNode)) {
            break;
        }
    }
   
    // so returns null
    return $returnNode;
}
```
A fix would be to pass through the original URL as well, and if it is on the tree, add the `$url` as a child to that node.

superpenguin612:
Fixes #466.

Could, in theory, pass only the original link to `addToDepthTree` instead, if a technically breaking change is undesirable.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -447,7 +447,7 @@ class Crawler
         }
     }
 
-    public function addToDepthTree(UriInterface $url, UriInterface $parentUrl, ?Node $node = null): ?Node
+    public function addToDepthTree(UriInterface $url, UriInterface $parentUrl, ?UriInterface $originalUrl = null, ?Node $node = null): ?Node
     {
         if (is_null($this->maximumDepth)) {
             return new Node((string) $url);
@@ -457,7 +457,7 @@ class Crawler
 
         $returnNode = null;
 
-        if ($node->getValue() === (string) $parentUrl) {
+        if ($node->getValue() === (string) $parentUrl || $node->getValue() === (string) $originalUrl) {
             $newNode = new Node((string) $url);
 
             $node->addChild($newNode);
@@ -466,7 +466,7 @@ class Crawler
         }
 
         foreach ($node->getChildren() as $currentNode) {
-            $returnNode = $this->addToDepthTree($url, $parentUrl, $currentNode);
+            $returnNode = $this->addToDepthTree($url, $parentUrl, $originalUrl, $currentNode);
 
             if (! is_null($returnNode)) {
                 break;
--- a/src/Handlers/CrawlRequestFulfilled.php
+++ b/src/Handlers/CrawlRequestFulfilled.php
@@ -62,8 +62,9 @@ class CrawlRequestFulfilled
         }
 
         $baseUrl = $this->getBaseUrl($response, $crawlUrl);
+        $originalUrl = $crawlUrl->url;
 
-        $this->urlParser->addFromHtml($body, $baseUrl);
+        $this->urlParser->addFromHtml($body, $baseUrl, $originalUrl);
 
         usleep($this->crawler->getDelayBetweenRequests());
     }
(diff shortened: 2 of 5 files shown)


freekmurze:
Is it possible to add a test for this to make sure the problem is fixed by this code?

superpenguin612:
The latest commit adds a test that fails on `main`, but succeeds on this branch.

src/Crawler.php
```diff
@@ -447,7 +447,7 @@ public function startCrawling(UriInterface|string $baseUrl)
         }
     }
 
-    public function addToDepthTree(UriInterface $url, UriInterface $parentUrl, ?Node $node = null): ?Node
+    public function addToDepthTree(UriInterface $url, UriInterface $parentUrl, ?UriInterface $originalUrl = null, ?Node $node = null): ?Node
```
freekmurze, line 450: Could you make this non-breaking by moving the new parameter to the last position?

freekmurze (changes requested):

src/Crawler.php
superpenguin612: Sure!

freekmurze:
The tests are failing. Could you take a look?

superpenguin612:
Sorry about that!

freekmurze:
Thanks!

PR #469, opened 2024-07-29
Catch unhandled Browsershot exceptions in `crawlFailed`

### LINKED ISSUE #324: Browsershot exceptions unhandled
I am using the following Browsershot configuration for the crawler:

```
$browsershot = new Browsershot();
$browsershot->waitForFunction("document.querySelector('#special_element') != null", 500, 5000);
```

Most pages on the site have the `#special_element`, but when the crawler encounters a page that does not have this element it exits with `Symfony\Component\Process\Exception\ProcessFailedException`. This is not caught in the `crawlFailed()` method of the `CrawlObserver`.

The trace provided from the command is as follows:

```
TimeoutError: waiting for function failed: timeout 5000ms exceeded
    at new WaitTask (/home/vagrant/sites/projectX/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:411:34)
    at DOMWorld.waitForFunction (/home/vagrant/sites/projectX/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:375:16)
    at Frame.waitForFunction (/home/vagrant/sites/projectX/node_modules/puppeteer/lib/cjs/puppeteer/common/FrameManager.js:910:32)
    at Page.waitForFunction (/home/vagrant/sites/projectX/node_modules/puppeteer/lib/cjs/puppeteer/common/Page.js:1328:33)
    at callChrome (/home/vagrant/sites/projectX/vendor/spatie/browsershot/bin/browser.js:235:24)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  name: 'TimeoutError'
}

  at vendor/spatie/browsershot/src/Browsershot.php:745
    741▕         if ($process->getExitCode() === 2) {
    742▕             throw new ElementNotFound($this->additionalOptions['selector']);
    743▕         }
    744▕
  ➜ 745▕         throw new ProcessFailedException($process);
    746▕     }
    747▕
    748▕     protected function getFullCommand(array $command)
    749▕     {
```

Is this something that the crawler should/could pass to `crawlFailed()` to handle? If not, how can I prevent Browsershot exceptions from halting the process of crawling the rest of the site?

superpenguin612:
Fixes #324.

I ran into this issue myself, and it ended up crashing the crawler prematurely. This PR catches any `ProcessFailedException`s on the Browsershot method, and sends it to observers via `crawlFailed`. I also included a test that loads a URL that never goes to network idle, triggering the 30s network timeout from Puppeteer. Originally, this creates an exception, but now it is caught appropriately.

One note here is that exceptions like Puppeteer not finding a Chrome binary are no longer displayed to the user by default, they would need to see the exception in `crawlFailed`. I'm not sure if there is a way to "fix" that, but that in theory should be what happens, anyway. 

Feel free to ask to use a different method of testing this; this test takes 30s to execute which may be undesirable (maybe use Browsershot's `waitForFunction` to get the 5s timeout, like what @kmcluckie posted)?

### DIFF
--- a/src/Handlers/CrawlRequestFulfilled.php
+++ b/src/Handlers/CrawlRequestFulfilled.php
@@ -15,6 +15,9 @@ use Spatie\Crawler\CrawlProfiles\CrawlSubdomains;
 use Spatie\Crawler\CrawlUrl;
 use Spatie\Crawler\ResponseWithCachedBody;
 use Spatie\Crawler\UrlParsers\UrlParser;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Exception\RequestException;
 
 class CrawlRequestFulfilled
 {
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

Issue #369, opened 2021-07-14
Binary files are parsed, ignoring setParseableMimeTypes


jespejoh:
When setting both attributes, the crawler sometimes fails to ignore binary files (e.g. ZIP, video). I've pinned the error down to the __invoke() function on the CrawlRequestFulfilled: 

        $body = $this->getBody($response);    <-- At this point $body is an empty string so the script continues but it fails at later stage.

        $robots = new CrawlerRobots(
            $response->getHeaders(),
            $body,
            $this->crawler->mustRespectRobots()
        );

        $crawlUrl = $this->crawler->getCrawlQueue()->getUrlById($index);

        if ($this->crawler->mayExecuteJavaScript()) {
            $body = $this->getBodyAfterExecutingJavaScript($crawlUrl->url);

            $response = $response->withBody(stream_for($body));
        }

I've managed to solve this issue locally by just adding this at the beginning of the function, so the crawling for this URL stops if the mime type is not supported:

        $contentType = $response->getHeaderLine('Content-Type');  

        if (! $this->isMimetypeAllowedToParse($contentType)) {
            return '';
        }

Not sure if it has any unexpected consequences but so far it works as expected. What do you think?

Happy to open a PR if you prefer it that way.

dominikfomicz:
have same error

PR #470, opened 2024-07-30
Fix setParsableMimeTypes()

superpenguin612:
Addresses issues #352 and #369, and PR #432.

The `setParsableMimeTypes` option does not actually prevent a *crawl* of a URL, it only prevents URLs that originate from that page from being crawled.

If `executeJavascript` is enabled in tandem, the `setParsableMimeTypes` option becomes completely broken; as it turns out, the method with which mime types are ignored conflicts with Browsershot's JavaScript execution, since it overwrites the `$body` variable:

```php
if ($this->crawler->mayExecuteJavaScript()) {
    $body = $this->getBodyAfterExecutingJavaScript($crawlUrl->url);
    $response = $response->withBody(Utils::streamFor($body));
}
```

This PR prevents pages with non-parsable mime types from being crawled completely, edits the test for mime types to be more representative of what this feature should accomplish, and adds an additional test to validate that the issue is fixed when executing JS, as well.

### DIFF
--- a/src/Handlers/CrawlRequestFulfilled.php
+++ b/src/Handlers/CrawlRequestFulfilled.php
@@ -29,6 +29,10 @@ class CrawlRequestFulfilled
     public function __invoke(ResponseInterface $response, $index)
     {
         $body = $this->getBody($response);
+        if (empty($body)) {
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #480, opened 2024-12-11
Add execution time limit

VincentLanglet:


VincentLanglet:
> Nice, could you maybe also provide a way that user can check if the time limit has been hit. Maybe a function `onTimeLimitHit($callable)` That passed callable could be called when the time limit has been hit.

I'm not sure to understand the request. What would be used for ?

This might add some complexity especially because you can hit time limit
- Because of the total time limit
- Because of the current time limit

I didn't see a `onCrawLimitHit()` methods...

Or maybe you would prefer that I split 
```
public function reachedCrawlLimits(): bool
```
into
```
public function reachedCrawlLimits(): bool
public function reachedTimeLimits(): bool
````
?

This way these method will be able to be called publicly to know if the limit is reached.

> Could you also update the docs?

Sure, I wanted to get a feedback from the Ci, since I didn't succeed running test on my computer...

VincentLanglet:
@freekmurze Ci is not triggered, not sure if it requires an action from you ?

freekmurze:
CI running now

VincentLanglet:
> CI running now

Thanks, tests are green

freekmurze:
Thanks!

PR #484, opened 2025-02-24
set spatie/browsershot minimal version to 5.0.5

grafst:
Because of a security vulnerability in the older versions: https://nvd.nist.gov/vuln/detail/CVE-2025-1022


freekmurze:
Thanks!

PR #485, opened 2025-05-03
Do not try robots.txt when ignored

kissifrot:
At the moment, whenever we start crawling, a call is made to the being-crawled server to retrieve its `robots.txt` file.
Problem is, when server does not have a `robots.txt`, is forbidden or anything, it creates noise and can even lead to a ban.

This PR ensures we won't try to retrieve robots.txt file when it was ignored.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -294,7 +294,7 @@ class Crawler
         return $this->rejectNofollowLinks;
     }
 
-    public function getRobotsTxt(): RobotsTxt
+    public function getRobotsTxt(): ?RobotsTxt
     {
         return $this->robotsTxt;
     }
@@ -478,11 +478,11 @@ class Crawler
 
         $crawlUrl = CrawlUrl::create($this->baseUrl);
 
-        $this->robotsTxt = $this->createRobotsTxt($crawlUrl->url);
+        if ($this->respectRobots) {
+            $this->robotsTxt = $this->createRobotsTxt($crawlUrl->url);
+        }
 
-        if ($this->robotsTxt->allows((string) $crawlUrl->url, $this->getUserAgent()) ||
-            ! $this->respectRobots
-        ) {
+        if ($this->shouldAddToCrawlQueue($crawlUrl)) {
             $this->addToCrawlQueue($crawlUrl);
         }
 
@@ -525,6 +525,23 @@ class Crawler
         }
 
         return $returnNode;
+    }
+
+    protected function shouldAddToCrawlQueue($crawlUrl): bool
+    {
+        if (!$this->respectRobots) {
+            return true;
+        }
+
+        if ($this->robotsTxt === null) {
+            return false;
+        }
+
+        if ($this->robotsTxt->allows((string) $crawlUrl->url, $this->getUserAgent())) {
+            return true;
+        }
+
+        return false;
     }
 
     protected function startCrawlingQueue(): void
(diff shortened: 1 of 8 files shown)


freekmurze:
Could you add a tests for this to ensure that the code works as intended?

kissifrot:
I had to update a few tests to make them work with GH actions, I hope you don't mind :sweat_smile:

kissifrot:
@freekmurze Looks like there was a dependency issue with Pest, it should run okay now :crossed_fingers:

kissifrot:
Up please :pray:

src/Crawler.php
```diff
@@ -478,9 +478,11 @@ public function startCrawling(UriInterface|string $baseUrl)
 
         $crawlUrl = CrawlUrl::create($this->baseUrl);
 
-        $this->robotsTxt = $this->createRobotsTxt($crawlUrl->url);
+        if ($this->respectRobots) {
+            $this->robotsTxt = $this->createRobotsTxt($crawlUrl->url);
+        }
 
-        if ($this->robotsTxt->allows((string) $crawlUrl->url, $this->getUserAgent()) ||
+        if (null !== $this->robotsTxt && $this->robotsTxt->allows((string) $crawlUrl->url, $this->getUserAgent()) ||
```
freekmurze, line 485: To me, this condition is unreadable (and not easy to debug). Could you extract it to a function with a good name, and use early returns there to avoid using `&&` and `||`.

Also don't use Yoda style comparisons, but regular ones.

src/UrlParsers/LinkUrlParser.php
```diff
@@ -87,7 +87,7 @@ protected function normalizeUrl(UriInterface $url): UriInterface
 
     protected function shouldCrawl(Node $node): bool
     {
-        if ($this->crawler->mustRespectRobots() && ! $this->crawler->getRobotsTxt()->allows($node->getValue(), $this->crawler->getUserAgent())) {
+        if ($this->crawler->mustRespectRobots() && null !== $this->crawler->getRobotsTxt() && ! $this->crawler->getRobotsTxt()->allows($node->getValue(), $this->crawler->getUserAgent())) {
```
freekmurze, line 90: Could you break this condition up, use early returns, and avoid `&&` and `||`

src/UrlParsers/SitemapUrlParser.php
```diff
@@ -74,7 +74,7 @@ protected function normalizeUrl(UriInterface $url): UriInterface
 
     protected function shouldCrawl(Node $node): bool
     {
-        if ($this->crawler->mustRespectRobots() && ! $this->crawler->getRobotsTxt()->allows($node->getValue(), $this->crawler->getUserAgent())) {
+        if ($this->crawler->mustRespectRobots() && null !== $this->crawler->getRobotsTxt() && ! $this->crawler->getRobotsTxt()->allows($node->getValue(), $this->crawler->getUserAgent())) {
```
freekmurze, line 77: Could you break this condition up, use early returns, and avoid `&&` and `||`

freekmurze (changes requested):

src/Crawler.php
kissifrot: Done :+1:

src/UrlParsers/LinkUrlParser.php
kissifrot: Done :+1:

src/UrlParsers/SitemapUrlParser.php
kissifrot: Done :+1:

freekmurze:
Thanks!

PR #504, opened 2026-03-28
Add a shouldStopCallback hook for graceful external stops

kissifrot:
## Add a shouldStopCallback hook for graceful external stops
 
- Some applications need to stop an active crawl based on external state, such as a database flag, an admin action, or a feature toggle.
- Today, the only graceful stop mechanism is the internal `shouldStop` flag, which makes these integrations awkward and pushes tests toward mutating internal state.
- This adds a fluent `shouldStopCallback()` hook that is evaluated before scheduling each next request. When it returns `true`, the crawl stops gracefully and `start()` returns `FinishReason::Interrupted`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,23 @@ Crawler::create('https://example.com')
     ->foundUrls();
 ```
 
+If you need to stop a crawl based on external state, you can register a callback that receives the current crawler instance and is checked before scheduling each next request:
+
+```php
+use Spatie\Crawler\Crawler;
+
+$shouldStop = false;
+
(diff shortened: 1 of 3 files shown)


kissifrot:
Little ping @freekmurze :pray:

freekmurze (approved):
Clean implementation that follows the existing fluent API patterns. Good phpdoc, the callback is only invoked when set (no perf impact for existing users), and test coverage covers both immediate and mid-crawl stops. CI is green across all matrix jobs.

PR #508, opened 2026-06-12
Skip signal handler registration when pcntl functions are disabled

poldixd:
This merge request fixes an issue on shared hosting.

The crawler checked if the pcntl extension is loaded, but on some shared servers the extension is loaded while the needed functions are disabled.

Because of that, the crawler tried to call pcntl_async_signals() and failed with an error.

### DIFF
--- a/src/Crawler.php
+++ b/src/Crawler.php
@@ -563,7 +563,11 @@ class Crawler
 
     protected function registerSignalHandlers(): void
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks for the fix, @poldixd! Good catch on shared hosting where `pcntl` is loaded but its functions are disabled. Released in 9.3.2.

PR #510, opened 2026-07-21
Fix Guzzle 8 exception handling for responses and retries

GrahamCampbell:
This follows up on #509, which added Guzzle 8 support but did not properly account for its exception reclassification, introducing a bug. Guzzle 8 moved `getResponse()` off the base `RequestException` onto its response-carrying subclasses, so `CollectUrlsObserver` reading it only from `BadResponseException` and `TooManyRedirectsException` misses the other response-carrying types on Guzzle 8 and drops responses that Guzzle 7 exposed on a plain `RequestException`. It now guards with `method_exists()` and handles every response-carrying exception properly.

Separately, Guzzle 8 reclassified read and operation timeouts from `ConnectException` to `NetworkTimeoutException`/`NetworkException`, which are siblings of `ConnectException` rather than subclasses, so the retry middleware stopped retrying the most common transient failure even though Guzzle 7 retried it. It now also matches `Psr\Http\Client\NetworkExceptionInterface`, which covers the whole network-failure family on both Guzzle 7 and 8.

On the dependency side, the `guzzlehttp/guzzle` and `guzzlehttp/psr7` constraints are raised to `^7.8.2 || ^8.0` and `^2.6.3 || ^3.0`, the first releases that support the PHP 8.4, and `psr/http-client` is now required directly since the retry change relies on its `NetworkExceptionInterface`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,8 @@
     ],
     "require": {
         "php": "^8.4",
-        "guzzlehttp/guzzle": "^7.3|^8.0",
-        "guzzlehttp/psr7": "^2.0|^3.0",
+        "guzzlehttp/guzzle": "^7.8.2 || ^8.0",
+        "guzzlehttp/psr7": "^2.6.3 || ^3.0",
         "psr/http-client": "^1.0",
         "spatie/robots-txt": "^2.0",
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you!

PR #512, opened 2026-08-07
Fix SitemapUrlParser extracting no URLs from namespaced sitemaps

### LINKED ISSUE #511: SitemapUrlParser extracts zero URLs from sitemaps that declare a second XML namespace (e.g. xmlns:image)
### Summary

`SitemapUrlParser` selects `<loc>` nodes with the unprefixed XPath `//loc`. That expression matches **zero** nodes whenever the sitemap's root element declares any prefixed namespace in addition to the default sitemap namespace — for example the very common Google image extension:

```xml
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
        xmlns:image="http://www.google.com/schemas/sitemap-image/1.1">
```

The crawl silently completes having found nothing. No exception, no warning — it just looks like an empty sitemap.

### Why it works for some sitemaps and not others

This is not a `symfony/dom-crawler` bug, and it is not version-dependent. `Crawler::addXmlContent()` deliberately strips the default namespace, but **only when it is the sole namespace in the document**:

```php
// symfony/dom-crawler — Crawler::addXmlContent()
// remove the default namespace if it's the only namespace to make XPath expressions simpler
if (!str_contains($content, 'xmlns:')) {
    $content = str_replace('xmlns', 'ns', $content);
}
```

So:

| Sitemap root | `xmlns:` present? | Default ns stripped? | `//loc` matches |
|---|---|---|---|
| `<urlset xmlns="…sitemap/0.9">` | no | yes | ✅ works |
| `<urlset xmlns="…sitemap/0.9" xmlns:image="…">` | yes | no | ❌ **0 nodes** |
| `<urlset>` (no namespace at all) | no | n/a | ✅ works |

`//loc` therefore only ever worked on the subset of sitemaps that declare exactly one namespace. The existing test fixtures in `tests/Pest.php` all fall into that subset, which is why the suite is green.

### Reproduction

```php
use Spatie\Crawler\UrlParsers\SitemapUrlParser;

$xml = <<<'XML'
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
        xmlns:image="http://www.google.com/schemas/sitemap-image/1.1">
   <url><loc>https://example.com/a</loc></url>
   <url><loc>https://example.com/b</loc></url>
</urlset>
XML;

count((new SitemapUrlParser)->extractUrls($xml, 'https://example.com'));
// expected: 2
// actual:   0
```

Remove `xmlns:image` from the root element and the same input returns `2`.

### Expected vs actual

| | |
|---|---|
| **Expected** | Every `<loc>` under `<url>` / `<sitemap>` is extracted, regardless of what namespaces the root declares. |
| **Actual** | Zero URLs extracted as soon as any `xmlns:*` prefix is declared on the root. |

### Real-world impact

Namespace extensions are the norm, not the exception — `xmlns:image`, `xmlns:video`, `xmlns:news`, and `xmlns:xhtml` are emitted by default by most sitemap generators (Drupal's Simple XML Sitemap, Yoast, RankMath, Screaming Frog, and others).

Concrete example: `https://www.ironman.com/sitemap.xml?page=2` (generated by Drupal's Simple XML Sitemap, declares `xmlns:image`) contains **2,000** `<loc>` entries. `SitemapUrlParser` returns **0**.

Because the failure is silent, `->parseSitemaps()` appears to work while quietly crawling nothing.

### Affected versions

Reproduced on `main` (`ef28855`, the current 9.4.1 line) with PHP 8.4.19 and `symfony/dom-crawler` v8.1.1, and on 8.4.3 with `symfony/dom-crawler` v7.4.4. The `//loc` expression is identical on `main`, on the `v9` branch and on the 8.x line, and the `addXmlContent()` behaviour it depends on is long-standing in dom-crawler, so this affects every released version that ships `SitemapUrlParser`.

### Relationship to #474

[#474](https://github.com/spatie/crawler/issues/474) ("For sitemap crawl url is not working (url loc from inside sitemap are not crawled)") reported the same user-visible symptom. It included the calling code but no sitemap sample and no diagnosis, and was closed with no comments. This issue supplies the missing part: the trigger is the presence of a prefixed namespace on the sitemap root, and the cause is the unprefixed `//loc` in `SitemapUrlParser::extractUrls()`. It is plausible that #474 was this bug.

### Suggested fix

Match on `local-name()` so the expression is namespace-agnostic, scoped to the parent elements the sitemap spec allows so that `<image:loc>` asset URLs are not swept up as pages:

```php
'//*[local-name()="url" or local-name()="sitemap"]/*[local-name()="loc"]'
```

Note that registering the sitemap namespace and querying `//sitemap:loc` is **not** a workable alternative — it would fix the multi-namespace case but break the single-namespace and no-namespace cases, since dom-crawler has already stripped the namespace in exactly those documents.

Happy to open a PR with the fix and regression tests.

witheez:
Fixes #511.

### The problem

`SitemapUrlParser::extractUrls()` selected `<loc>` nodes with the unprefixed XPath `//loc`. That matches zero nodes as soon as the sitemap root declares any prefixed namespace alongside the default one:

```xml
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
        xmlns:image="http://www.google.com/schemas/sitemap-image/1.1">
```

`symfony/dom-crawler` strips the default namespace in `addXmlContent()`, but only when it is the *only* namespace in the document:

```php
// remove the default namespace if it's the only namespace to make XPath expressions simpler
if (!str_contains($content, 'xmlns:')) {
    $content = str_replace('xmlns', 'ns', $content);
}
```

So `//loc` worked only for sitemaps declaring exactly one namespace. Any sitemap using the image / video / news / xhtml extensions — which is most of them in practice — returned zero URLs, silently.

### The change

One line in `src/UrlParsers/SitemapUrlParser.php`:

```diff
-        $domCrawler->filterXPath('//loc')->each(function (DomCrawler $node) use (&$urls, &$seen) {
+        $locXPath = '//*[local-name()="url" or local-name()="sitemap"]/*[local-name()="loc"]';
+
+        $domCrawler->filterXPath($locXPath)->each(function (DomCrawler $node) use (&$urls, &$seen) {
```

Matching on `local-name()` makes the selection namespace-agnostic, so it behaves identically whether or not dom-crawler stripped the namespace.

**On the two design choices:**

1. **Why not register the namespace and query `//sitemap:loc`?** Because it fixes the broken case and breaks the working ones. Verified against all three shapes:

   | Sitemap | `registerNamespace` + `//sitemap:loc` | `local-name()` |
   |---|---|---|
   | single-namespace `<urlset xmlns="…">` | ❌ 0 | ✅ 1 |
   | multi-namespace `<urlset xmlns="…" xmlns:image="…">` | ✅ 1 | ✅ 1 |
   | no namespace `<urlset>` | ❌ 0 | ✅ 1 |

   dom-crawler has already removed the namespace in the first and third rows, so there is nothing left to match a prefix against.

2. **Why scope to `url` / `sitemap` parents instead of a plain `//*[local-name()="loc"]`?** Because the bare version also matches `<image:loc>`, which holds an image asset URL rather than a page, and would start feeding image binaries into the crawl queue:

   | Fixture | `//loc` (before) | `//*[local-name()="loc"]` | scoped (this PR) |
   |---|---|---|---|
   | single-ns urlset | 1 | 1 | 1 |
   | multi-ns urlset | 0 | 1 | 1 |
   | multi-ns + `<image:loc>` | 0 | **2** ← image swept in | 1 |
   | multi-ns sitemapindex | 0 | 1 | 1 |
   | ironman.com sitemap (2,000 entries, real) | **0** | 2000 | 2000 |

   The sitemap protocol only permits `<loc>` as a child of `<url>` (in a `urlset`) or `<sitemap>` (in a `sitemapindex`), so the scoping matches the spec. If you'd prefer the shorter bare form and consider the image behaviour acceptable, that's a one-word change — happy to switch.

### Tests

Two tests added to `tests/UrlParsers/SitemapUrlParserTest.php`, plus a `namespacedSitemap()` fixture and two fakes in `tests/Pest.php`, following the existing integration-through-the-crawler style:

- **`it should extract urls from a sitemap that declares additional namespaces`** — fails on `main` (0 URLs crawled), passes with the fix.
- **`it should not extract image urls from a sitemap using the image extension`** — passes with the fix, fails if the XPath is loosened to the bare `//*[local-name()="loc"]`.

Both were confirmed to be load-bearing by mutating the XPath in three directions and checking that the intended test fails each time:

| `$locXPath` value | namespaces test | image test |
|---|---|---|
| `//loc` (original) | ❌ fails | ✅ passes |
| `//*[local-name()="loc"]` | ✅ passes | ❌ fails |
| scoped (this PR) | ✅ passes | ✅ passes |

The existing fixtures all declared a single namespace, which is why the suite was green with the bug present.

### Verification

- `composer test` — 229 passed (552 assertions), whole suite including `tests/Integration`.
- `composer analyse` — no errors.
- Pint (default Laravel preset) — clean on the three changed files.
- Against a real-world 2,000-entry sitemap (`ironman.com`, Drupal Simple XML Sitemap, declares `xmlns:image`): 0 URLs before, 2,000 after.

### Backwards compatibility

No public API change. Behaviour is identical for every sitemap that worked before; sitemaps that previously yielded nothing now yield their URLs. The one narrow behavioural difference is that a `<loc>` placed somewhere other than under `<url>` / `<sitemap>` is no longer picked up — such a document isn't valid per the sitemap protocol.


freekmurze:
Thanks for the thorough writeup and tests, @witheez! Merged and released in 9.4.2.

